PR #1, opened 2016-01-11
CORE-1565 HTTP API docs

fmitchell-r7:
This is a first pass at documenting what the HTTP API looks like from an user perspective.


athompson-r7:
:+1:

PR #6, opened 2016-01-19
Core-1560 parser documentation

rvanburen-r7:
Initial parser.md file


docs/class-parser.md
```diff
@@ -0,0 +1,34 @@
...
+
+### Class Attribute: `sources`
+
+```
+Parser.sources
+```
+
+The source of the properties. ie S3, Conqueso, ect
+
+### Constructor(properties (buffer), plugin_type (string))
```
fmitchell-r7: Where would `plugin_type` be used within this class?

rvanburen-r7: It would be used because different sources may give different types of buffers. So S3 source translates YMAL but the Local Source translates XML instead. this way its one call but can handle a bunch of different property languages.

dgreene-r7: To me, this raises the question as to whether we want a monolithic parser that parses all expected input or if we want discrete parsers for every data source. Per a conversation with @fmitchell-r7 I'm leaning more towards the latter. It fits a little better into node's paradigm of using small, discrete packages and if we get to a point where we're duplicating code, we can think about extracting it into separate helper packages.

rvanburen-r7: I was under the same impression John was. 
However, after the meeting yesterday, this documentation will be updated to reflect the new model.

docs/sources/s3/class-source-s3-parser.md
```diff
@@ -0,0 +1,9 @@
+## Sources.S3.Parser
+Parser is the class that will translate strings (starting with yamal) into properties (json) between storage and services
```
dgreene-r7: In this case, no YAML will be involved. This parser will just be translating data pulled from the aws-sdk into our storage layer's json schema.

dgreene-r7:
My preference for parsers is a very opaque interface. You should have a `constructor`, whatever internals you need to parse the data, and a `getData` method or something like that. 

If I'm using a parser all I want is to create a new one with my data and get back the parsed data when I ask for it.

PR #8, opened 2016-01-27
Added additional documentation for a source interface

dgreene-r7:


docs/sources/source-interface.md
```diff
@@ -0,0 +1,28 @@
...
+## Methods
+1. A constructor that accepts a path to a given resource. That resource can be local or remote as long as the plugin itself understands how to parse the path and retrieve the data. Additional arguments are acceptable as long as the index document contains information to fill in the remaining arguments.
+1. `getName()`: (`string`) Returns a unique name for the plugin instance.
+	
+	In the S3 source, for example, `getName()` returns a string comprised of the plugin type, the resource bucket, and the resource path.
+	
+1. `getType()`: (`string`) Returns the plugin type.
+1. `fetch(callback, args)` Executes the callback with the given `args`. 
+
+	Built-in plugins have a `defaultFetch(options)` method that provides a default callback. For instance, the S3 plugin can be executed simply by calling `s3.fetch()`.
```
fmitchell-r7: If I'm writing my own plugin, do I need a `defaultFetch` method, or can I simply define `fetch`?

dgreene-r7: I think `fetch` is the only one that's necessary. My goal with using `defaultFetch` was to make the plugin work exactly how you would think (go get a file, parse the file, emit event) but expose `fetch` to allow users to override the default behavior within the plugin context.

rvanburen-r7:
:shipit:

docs/sources/source-interface.md
dgreene-r7: See https://github.com/rapid7/propsd/blob/CORE-1613_s3_source_plugin/src/sources/s3/s3.js#L46-L88 for an example.

PR #9, opened 2016-01-29
CORE-1613: First pass at a S3 plugin implementation.

dgreene-r7:


lib/sources/s3/agent.js
```diff
@@ -0,0 +1,81 @@
...
+    _classCallCheck(this, Agent);
+
+    this._bucket = bucket;
+    this._path = path;
+    this._s3 = new AWS.S3();
+  }
+
+  /**
+   * Generates query params for S3
+   * @param eTag
```
fmitchell-r7: Is the `eTag` parameter a string, or does the SDK map it to some kind of structure?

lib/sources/s3/agent.js
```diff
@@ -0,0 +1,81 @@
...
+
+      if (eTag) {
+        s3params.IfNoneMatch = eTag;
+      }
+      return s3params;
+    }
+
+    /**
+     * Fetches an object from S3
+     * @param eTag
```
fmitchell-r7: Is the `eTag` parameter a string, or does the SDK map it to some kind of structure?

dgreene-r7: Just a string.

lib/sources/s3/agent.js
dgreene-r7: Just a string.

src/sources/s3/s3.js
```diff
@@ -0,0 +1,105 @@
...
+   */
+  status() {
+
+  }
+
+  /**
+   * Shutdown method
+   */
+  shutdown() {
+    this._timer.clear();
```
fmitchell-r7: I don't think there's a `clear()` function on `intervalObjects`. Should this be `clearInterval(this._timer);` instead?

PR #14, opened 2016-02-05
Add module tests

jmanero-r7:
Mostly stubs ATM.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,15 +1,18 @@
 {
   "name": "propsd",
   "version": "0.0.1",
-  "description": "A local daemon for loading static and dynamic properties",
+  "author": "Rapid7 <coreservices@rapid7.com>",
+  "contributors": [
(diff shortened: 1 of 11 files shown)


jmanero-r7:
:/ Lots of lint issues that get fixed in the `es6` branch

package.json
```diff
@@ -1,15 +1,18 @@
...
   "main": "lib/index.js",
-  "directories": {
-    "doc": "docs",
-    "bin": "bin"
-  },
   "scripts": {
-    "test": "mocha",
-    "lint": "eslint lib/**"
+    "test": "./node_modules/.bin/mocha",
+    "lint": "./node_modules/.bin/eslint bin lib test"
```
dgreene-r7: You don't need to specify `./node_modules/.bin` here. When you execute `npm run *` it executes with `node_modules/.bin` added to the $PATH.

PR #15, opened 2016-02-05
1.0 Implementation Proposal

jmanero-r7:

PR #16, opened 2016-02-12
Feature Conqueso API

fmitchell-r7:
This sets up the [Conqueso compatibility layer](https://github.com/rapid7/propsd/blob/feature-conqueso-api/docs/http-api.md#conqueso) for the HTTP API.

The `Conqueso` function assumes it's been provided an instance of the `Storage` class so it has properties to render. My current assumption is that the Storage class takes care of flattening properties, so accessing the `.properties` value returns a POJO of key value pairs.

The HTTP `OPTIONS` method is now supported and the documentation's been updated to reflect that. I'm contemplating supporting `HEAD` as well (since we support `GET`). Though I don't think that's necessary for Conqueso compatibility.

### DIFF
--- a/lib/control/v1/conqueso.js
+++ b/lib/control/v1/conqueso.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const HTTP_OK = 200;
+const HTTP_METHOD_NOT_ALLOWED = 405;
 
 /**
  * Conqueso compatible API
@@ -14,6 +15,12 @@ function Conqueso(app) {
 
   app.put('/v1/conqueso*', (req, res) => {
     res.status(HTTP_OK).end();
+  });
+
+  app.delete('/v1/conqueso*', (req, res) => {
+    res.set('Allow', 'POST,PUT');
+    res.status(HTTP_METHOD_NOT_ALLOWED);
+    res.end();
   });
 }
 
--- a/lib/control/v1/conqueso.js
+++ b/lib/control/v1/conqueso.js
@@ -18,13 +18,13 @@ function Conqueso(app) {
   });
 
   app.options('/v1/conqueso*', (req, res) => {
-    res.set('Allow', 'POST,PUT');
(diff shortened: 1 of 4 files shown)


dgreene-r7:
2 things:

1. I assume the goal here is to backfill the rest of the app from @jmanero-r7's 1.0-proposal branch once linting has been fixed.

2. I thought we were doing ES6, no?

lib/control/v1/conqueso.js
```diff
@@ -0,0 +1,70 @@
...
+  route.head(methodNotAllowed);
+
+  route.post((req, res) => {
+    res.status(HTTP_OK);
+    res.end();
+  });
+
+  route.put((req, res) => {
+    res.status(HTTP_OK);
+    res.end();
```
dgreene-r7: We're ok with returning a `200` for PUT and POST? My instinct is to reject any methods that don't do anything out of hand.

dgreene-r7:
Also :+1: for tests! :grinning:

lib/control/v1/conqueso.js
fmitchell-r7: In terms of keeping Conqueso compatibility, I think this makes sense. Conqueso users expect to be able to create and set properties with PUT and POST requests. I don't want to break that use case, even though we're not doing anything with the data sent.

fmitchell-r7:
@dgreene-r7 "Yes" to ES6! Is there an ES6 ism I missed? I've got ES5 muscle memory, so I'm not surprised if I missed something.

fmitchell-r7:
@dgreene-r7 My goal was to make this compatible with @jmanero-r7's [1.0 proposal](https://github.com/rapid7/propsd/pull/15). There will be some additional code needed to wire up the Conqueso API once that's merged.

lib/control/v1/conqueso.js
dgreene-r7: :+1:

dgreene-r7:
:+1:

PR #17, opened 2016-02-17
Feature Core API

dgreene-r7:
Implemented the Core API which includes `/v1/health` and `/v1/status` endpoints. The `Core` function assumes it's been provided an instance of the `Storage` class so it has properties to render.


fmitchell-r7:
:+1: Looks good to me.

PR #18, opened 2016-02-17
Implemented logger

dgreene-r7:
The test will fail because it's tightly coupled with a config object that has an [nconf][]-based interface. This PR shouldn't be merged until the `config` object is implemented.

[nconf]: https://github.com/indexzero/nconf

### DIFF
--- a/test/logger.js
+++ b/test/logger.js
@@ -1,0 +1,12 @@
+/* eslint-env mocha */
+'use strict';
+const Winston = require('winston');
+
+describe('logging', () => {
+  it('returns a WINSTON object', () => {
+    const config = require('../lib/config');
+    const log = require('../lib/logger').attach(config);
+
+    log.should.be.an.instanceOf(Winston.Logger);
+  });
+});
--- a/test/logger.js
+++ b/test/logger.js
@@ -1,12 +1,44 @@
 /* eslint-env mocha */
 'use strict';
+
+require('should');
+const fs = require('fs');
 const Winston = require('winston');
 
-describe('logging', () => {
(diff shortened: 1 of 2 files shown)


test/logger.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-env mocha */
+'use strict';
+const Winston = require('winston');
+
+describe('logging', () => {
+  it('returns a WINSTON object', () => {
+    const config = require('../lib/config');
+    const log = require('../lib/logger').attach(config);
+
+    log.should.be.an.instanceOf(Winston.Logger);
```
fmitchell-r7: Instead of testing the logger type, we can test logger behavior.

Pass in a `Config` compatible object that logs to a temp file and check the following:
* That the logger defines convenience methods for `info`, `warn`, and `error` logging
* That logged messages show up in the temp file
* If the log level is set to `error`, messages logged at `info` and `warn` level don't show up

dgreene-r7: Not sure I agree unless we're saying that we don't care what logging lib we choose as long as it has those messages. To me, what you're suggesting is effectively the same as testing Winston.

However, thinking about it, having a global logger might not be a bad idea and I'm definitely going to build a `Config`-like object to pass into the logging module for testing.

fmitchell-r7: I feel like we shouldn't care what logging library we use, as long as it has the functionality we expect. The tests describe that functionality.

This may be something where we want to add tests to cover functionality as we use it, rather than trying to think through all the possible use case of a logging library up front.

PR #19, opened 2016-02-18
Added nvmrc file

dgreene-r7:
Let's keep our node versions consistent!


fmitchell-r7:
:+1: Thanks for adding this! I assume Travis CI will pick up on it as well?

dgreene-r7:
Per https://github.com/travis-ci/travis-build/pull/469 it loads the `.travis.yml` file first, then the `.nvmrc` and then falls back to the stable version of nvm. 

I'm wonder if we should shelve this PR, then, as we're "targeting" `5.5.x` and `4.2.x`. Travis lets us run builds against both of those rather than just one listed in the `.nvmrc`. 

Maybe we fix the `.nvmrc` file to `5.5.x` that way when we're developing we're working against the newest supported version that we're building against.

What do you think?

fmitchell-r7:
Let's put `5.5.x` in the `.nvmrc` file, since that's what CI is checking against. Keeping development and CI versions in sync makes it easier to run down broken builds.

dgreene-r7:
Agree. I'm going to yank build targets out of the `.travis.yml` file so the `.nvmrc` file is the canonical source of truth.

PR #26, opened 2016-02-19
Feature storage engine

fmitchell-r7:
This provides a `Storage` class to manage the collected set of merged properties from plugins. The `Storage#register` function is called to register plugins. When a plugin has updated properties, calling `Storage#update` to let the engine know new properties are available.

The responsibility for registering plugins with the `Storage` class and notifying the `Storage` class when plugins have new properties, is left to up an as of yet undefined `PluginManager`.

### DIFF
--- a/docs/class-storage.md
+++ b/docs/class-storage.md
@@ -7,10 +7,6 @@ _In a future version, the Storage module should be able to discover new sources 
 
 ## Interface
 
-### Constructor
-
-The Storage class can be instantiated with an optional `EventEmitter`. If one's provided, "update" events will be emitted to it whenever merged properties are modified.
-
 ### Class Attribute: `properties`
 
 ```
--- a/docs/class-storage.md
+++ b/docs/class-storage.md
@@ -27,6 +27,10 @@ An array of active Source instances
 
 Sources should call the `update` method to notify the Storage class when they have new properties.
 
+### Instance Method: `register(source)`
+
+Sources should register with the Storage class by calling the `register` method. The `register` method takes the source as an argument. Only registered sources' properties will be merged when `update` is called.
+
 ## Class: Storage.Index
 
 A Scheme for defining an ordered set of sources to be managed by the Storage module. This scheme will parse a JSON array of Source definitions:
--- a/docs/class-storage.md
+++ b/docs/class-storage.md
@@ -3,9 +3,13 @@ Class: Storage
 
 The `Storage` module is responsible for managing data sources and merging their properties into a single document in the correct order. In "Version 1", it will use a dedicated S3Watcher to fetch an index object from S3, and add/remove sources accoding to an array defined therein.
 
-_In a future version, the Storage module should be able to discover new sources from fetched objects, removing the need for an out-of-band definition of sources in an index. Instead, the Storage module would initialize its first source(s) from local configuration, and find additional sources recursivly from their own contents._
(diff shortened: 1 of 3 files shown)


lib/storage.js
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * @return {Boolean} Returns true if the value will overwrite others, false otherwise.
+ */
+function canOverwriteWith(value) {
+  return value !== null && value !== undefined && !(value instanceof Function);
+}
+
+/**
+ * Perform a deep merge of objects based on key.
+ * Warning! Destination is modified in place.
```
dgreene-r7: :cold_sweat:

dgreene-r7:
This seems like it would solve your deep merging issues without modifying anything in place: http://facebook.github.io/immutable-js/docs/#/Map/mergeDeepWith

Plus immutable-js offers an `OrderedMap` and a bunch of other data structures we could make use of to simplify stuff.

lib/storage.js
```diff
@@ -0,0 +1,111 @@
...
+
+class Storage {
+  /**
+   * The storage engine maintains a merged set of properties retrived from
+   * source plugins. An optional EventEmitter can be provided. When properties
+   * change an "update" event will fire on the emitter with the new properites.
+   *
+   * @param {EventEmitter} emitter
+   */
+  constructor(emitter) {
```
dgreene-r7: Why accept an emitter as an arg instead of just extending `EventEmitter`? The `Storage` object seems like it will stick around for the lifetime of the process, so why not have other things subscribe to it rather than vice-versa?

lib/storage.js
fmitchell-r7: I haven't found a merge library that does what I want yet. For the object case, I want a deep merge based on keys. For _everything else_ I want to overwrite existing values. This mostly breaks downf or `Array`. Everything I've found seems to want to do

```javascript
merge(['a', 'b'], ['c']) === ['c', 'b'])
```

fmitchell-r7:
The mention of immutability does make me think I should do something like `JSON.parse(JSON.stringify())` after getting properties from a plugin and before before merging them.

lib/storage.js
fmitchell-r7: I wanted to allow for a global event bus that all the events in the system can flow through.

dgreene-r7: Wouldn't it make sense for the `Storage` object to be that event bus? It persists for the entire runtime and orchestrates most of the work. Seems like a perfect candidate to me.

fmitchell-r7: Maybe? I was following the Single Responsibility Principle and thinking that `Storage` should only be responsible for things related to properties. The event bus didn't feel related.

fmitchell-r7:
To capture an out of band conversation with @dgreene-r7 here... We're going to make `Storage#update` synchronous and move the responsibility of calling the update function to the module that's handling listening to source plugins for update events.

docs/class-storage.md
```diff
@@ -23,6 +23,14 @@ Storage.sources
 
 An array of active Source instances
 
+### Instance Method: `update()`
+
+Sources should call the `update` method to notify the Storage class when they have new properties.
```
dgreene-r7: Need to change the language here to capture how we're going to use `update()`.

test/storage-engine.js
```diff
@@ -0,0 +1,184 @@
...
+    storage.register(1.0);
+    storage.register('properties');
+    storage.register(() => {});
+    storage.register([]);
+    storage.register({});
+
+    should(storage.sources).eql([]);
+  });
+
+  it('merges properites on demand', () => {
```
dgreene-r7: Spelling

dgreene-r7:
This LGTM. I think you're good to merge it.

PR #27, opened 2016-02-22
PD added string-template feature from 1.0_proposal branch

rvanburen-r7:


lib/string-template.js
```diff
@@ -0,0 +1,56 @@
...
+      }
+
+      return node;
+    });
+  }
+}
+
+StringTemplate.CAPTURE = /\{\{ ?(.+?) ?\}\}/g;
+StringTemplate.DELIMITER = ':';
+StringTemplate.prototype.toJSON = StringTemplate.prototype.toString;
```
fmitchell-r7: Is there a test case for `toJSON`? I didn't see one, but I may have just missed it.

test/string-template.js
```diff
@@ -0,0 +1,45 @@
+/* eslint-env mocha */
+'use strict';
+
+const should = require('should');
+const StringTemplate = require('../lib/string-template');
+
+describe('StringTemplate', () => {
+  let string = 'The {{ jumper:adjective }} {{ jumper:color }} {{ jumper:animal }}';
```
fmitchell-r7: We might want to change one of these to account for the optional spaces in the capture group e.g. "{{jumper:adjective}}" instead of "{{ jumper:adjective }}".

lib/string-template.js
rvanburen-r7: I didn't test it as it is a built in function, but I am adding a test now.

test/string-template.js
rvanburen-r7: next commit will have both "{{jumper:adjective}}" and "{{ jumper:adjective }}".

fmitchell-r7:
:+1: Looks good to me.

PR #28, opened 2016-02-22
Fix logger without file

fmitchell-r7:
This fixes and issue where the logger would crash if you attempt to instantiate it without a config file.


dgreene-r7:
:+1:

I kept getting this error and was like "eh, I'll fix it later."

PR #29, opened 2016-02-22
Feature application server

fmitchell-r7:
This sets up a basic HTTP server for propsd. None of the underpinning storage or plugin layers are there yet. But logging works and the server responds to GET requests on the `/v1/health` endpoint.

Singleton global `Config` and `Log` objects are defined so they can be leveraged by plugins.


athompson-r7:
Nice readme.

Do the logs automatically roll or is that another set of config?

fmitchell-r7:
@athompson-r7 Winston does provide [file rotation based on size and max number of files](https://github.com/winstonjs/winston/blob/master/docs/transports.md#file-transport). We could add configuration options to support that. For now, I updated the README to indicate that propsd doesn't handle log file rotation.

athompson-r7:
:+1:

dgreene-r7:
Looks good. Might want to toss it in the `bin/` directory just for consistency.

Also despite the awful code smell, we really should have singleton instances of the logger and config and put them in the global namespace just for simplicity's sake.

PR #30, opened 2016-02-23
Metadata source plugin

dgreene-r7:
This is the Metadata source plugin cribbed from @jmanero-r7's work on the 1.0-proposal branch. It removes the layers of inheritance in favor of being a single, self-contained unit which will allows us to test in isolation and eventually (maybe?) package and release source plugins as node package.


lib/source/metadata.js
```diff
@@ -0,0 +1,504 @@
...
+    this._updated = null;
+
+    /**
+     * Initialize the metadata-service client
+     */
+    this.service = new Aws.MetadataService({
+      httpOptions: {
+        timeout: METADATA_TIMEOUT
+      },
+      host: Config.get('metadata:host')
```
fmitchell-r7: What happens when the `host` parameter is `null` or `undefined`?

dgreene-r7: The only possible case we have where that _might_ happen is if the `Config` object isn't registered globally. the `metadata:host` key has a default value that's set in the `Config` module.

test/metadata.js
```diff
@@ -0,0 +1,136 @@
...
+describe('Metadata source plugin', () => {
+  beforeEach(() => {
+    this.m = new Metadata();
+    this.m.service.host = '127.0.0.1:8080';
+  });
+
+  afterEach(() => {
+    const status = this.m.status();
+
+    if (status.running) {
```
fmitchell-r7: Is it not safe to call `shutdown` if the plugin isn't running?

test/metadata.js
```diff
@@ -0,0 +1,136 @@
...
+    });
+
+    this.m.initialize();
+  });
+
+  it('shuts down cleanly', (done) => {
+    this.m.on('shutdown', () => {
+      const status = this.m.status();
+
+      status.running.should.be.false; // eslint-disable-line no-unused-expressions
```
fmitchell-r7: You can avoid the eslint comment by writing this as
```javascript
status.running.should.be.false();
```

test/metadata.js
```diff
@@ -0,0 +1,136 @@
...
+        signature = null,
+        secondExecution = false;
+
+    this.m.on('update', () => {
+      instanceId = this.m.properties.instance['ami-id'];
+      signature = this.m.signature;
+
+      secondExecution = true;
+
+      // This is a terrible hack
```
fmitchell-r7: What is the terrible hack doing?

test/metadata.js
```diff
@@ -0,0 +1,136 @@
...
+  });
+
+  it('exposes an error when one occurs but continues running', (done) => {
+    server.stop();
+    this.m.on('error', (err) => {
+      const status = this.m.status();
+
+      err.code.should.equal('ECONNREFUSED');
+      status.ok.should.be.false; // eslint-disable-line no-unused-expressions
+      status.running.should.be.true; // eslint-disable-line no-unused-expressions
```
fmitchell-r7: You can avoid the eslint comment by rewriting these as
```javascript
status.ok.should.be.false();
status.running.should.be.true();
```

test/utils/test-metadata-server.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+  response.writeHead(code, {'Content-Length': body.length, 'Content-Type': 'text/plain'});
+  response.end(body);
+});
+
+const sockets = {};
+let nextSocketId = 0;
+
+server.on('listening', () => {
+  console.log(`listening for metadata requests at http://127.0.0.1:${DEFAULT_HTTP_PORT}`); // eslint-disable-line
```
fmitchell-r7: Why use console logging here instead of a default Winston instance?

lib/source/metadata.js
fmitchell-r7: You're right. If someone's overwriting the metadata host to force it through a proxy or something they probably know what they're doing.

test/metadata.js
dgreene-r7: This terrible hack exists because I couldn't get `sinon`'s fake clock working correctly. The workaround is to `initialize()` the plugin (it runs and gets data), then setting the `_timer` to false (turning off the timout), then reinitializing.

The "right" way to do this would be something like this:
```javascript
const sinon = require('sinon');

before(() => {
  clock = sinon. useFakeTimers();
});

it('doesn\'t update data if the Metadata Service document is the same', (done) => {
  const spy = sinon.spy();
  this.m.on('update', spy);

  setTimeout(function() {
    spy.should.be.called();
    spy.should.be.calledOnce();
    done();
  }, this.m.interval);
  

  this.m.initialize();
  clock.tick(this.m.interval - 50);

});

after(() => {
  clock.restore();
});
```

But I could never get that to work correctly. So instead I opted to add an informative event `no-update` and that hack.

Please let me know if you have other ideas!

test/utils/test-metadata-server.js
dgreene-r7: To be honest, it's a test util that's entirely disconnected from the rest of the app. I didn't see a need to pass in a Winston instance and didn't want to instantiate one inside the util. I also didn't want to presume that a global `Log` object would exist in this context.

test/metadata.js
dgreene-r7: Also, interestingly enough, there doesn't seem to be a way to test if an object is an instance of `timeoutObject` so the way I'm doing it is by testing if the non-inherited keys are consistent (see test on line 42.) Using `sinon`'s fake timers breaks that test.

fmitchell-r7:
:+1: Thanks for wrangling the tests around this!

PR #32, opened 2016-02-25
Fixed typo which leads to config files never being loaded

dgreene-r7:


fmitchell-r7:
:+1: Thanks for catching that!

PR #33, opened 2016-02-26
Added two more string-template tests

dgreene-r7:


fmitchell-r7:
:+1: Thanks for leveling up our code coverage numbers!

PR #34, opened 2016-02-26
First pass at S3 plugin converted from @jmanero-r7's work

dgreene-r7:


lib/source/s3.js
```diff
@@ -0,0 +1,284 @@
...
+
+  /**
+   * @param {Object} params
+   * @returns {Boolean}
+   */
+  configure(params) {
+    let changed = false;
+
+    changed = S3.setIfChanged(this, 'bucket', params.bucket || Config.get('index:bucket')) || changed;
+    changed = S3.setIfChanged(this, 'interval', params.interval || DEFAULT_INTERVAL) || false;
```
fmitchell-r7: Should this be `|| changed` at the end instead of `|| false` like the other two?

test/s3.js
```diff
@@ -0,0 +1,191 @@
...
+    getObjectStub = sinon.stub().callsArgWith(1, null, fakeResponse);
+    S3 = generateProxy(getObjectStub);
+
+    this.s3 = new S3({bucket: DEFAULT_BUCKET, path: 'foo.json'});
+    done();
+  });
+
+  it('throws an error if instantiated without bucket or path', () => {
+    should.throws(() => {
+      const s3 = new S3(); // eslint-disable-line no-unused-vars
```
fmitchell-r7: Can you do `new S3()` here and avoid the eslint comment?

test/s3.js
```diff
@@ -0,0 +1,191 @@
...
+    }, Error);
+
+    should.throws(() => {
+      const s3 = new S3({  // eslint-disable-line no-unused-vars
```
fmitchell-r7: Can you just have `new S3({bucket: DEFAULT_BUCKET});` here and avoid the eslint comment?

lib/source/s3.js
```diff
@@ -0,0 +1,284 @@
...
+    this.properties = {};
+  }
+
+  update(data) {
+    let properties = {};
+
+    try {
+      properties = JSON.parse(data.toString());
+    } catch (e) {
+      // We should have an error condition but tbd
```
fmitchell-r7: I didn't see a test case for this `catch` branch, but I might have just missed it.

dgreene-r7: I'm of two minds here. On the one hand, having an error condition and then testing for it is effectively testing the body that AWS sends back. 

On the other hand, it's not a terrible idea to test in isolation, and now that we can effectively stub `getObject` it can't hurt to do something here. 

I'm amenable either way.

lib/source/s3.js
dgreene-r7: Good catch.

test/s3.js
dgreene-r7: `new S3();` triggers `no-new` so it's kind of a wash.

test/s3.js
dgreene-r7: Same as above comment.

fmitchell-r7:
:+1: The AWS mock in the tests is really slick.

dgreene-r7:
Thanks! That's the result of a LOT of research last night. Sinon and proxyquire are not intuitive at all. I think I might yank that out into a `test-utils` module for re-use so we don't have to re-invent the wheel.

PR #35, opened 2016-02-26
Fix for the metadata source plugin not playing nice with the storage layer's register method

dgreene-r7:
```javascript
// Storage.register
register(plugin) {
  if (canMerge(plugin) && plugin.hasOwnProperty('properties')) {
    this.sources.push(plugin);
  }
}
```
The `Metadata` source can't be registered with the storage layer because getters aren't picked up by `hasOwnProperty()`. We could change the storage layer, but `Object.getOwnPropertyDescriptor()` is inconsistent when the underlying data comes from an object with a different `__proto__`.

It's easier and more consistent to have a properties attribute on the `Metadata` class rather than use a getter.

### DIFF
--- a/lib/source/metadata.js
+++ b/lib/source/metadata.js
@@ -198,6 +198,7 @@ class Metadata extends EventEmitter {
     this.name = options.name || 'source';
     this._ok = false;
     this._updated = null;
+    this.properties = {};
 
     /**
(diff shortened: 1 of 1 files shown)


rvanburen-r7:
Makes sense :+1:

PR #37, opened 2016-03-02
Fix metadata signature bug

dgreene-r7:
This fixes a bug where even if a metadata source plugin instance is shut down it still maintains the sha1 signature of the data which means `update` events won't be emitted if the plugin is started again unless the underlying data changes.

### DIFF
--- a/lib/source/metadata.js
+++ b/lib/source/metadata.js
@@ -327,6 +327,7 @@ class Metadata extends EventEmitter {
     }
 
     this._ok = false;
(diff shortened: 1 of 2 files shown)


fmitchell-r7:
Is there a way to test this bug so we can be sure it stays fixed in the future?

dgreene-r7:
Yeah, I'll write a new test.

fmitchell-r7:
:+1: Looks good to me.

PR #38, opened 2016-03-02
Expose type property in source interface

dgreene-r7:
This is the first pull request that will culminate in the `PluginManager` implementation. The `PluginManager` requires a defined `type` for each source in order to create an instance of the correct source plugin from the index.

This should be an essentially static property per source class. The value of `type` should be reflected in the `_registerSources()` method in the `PluginManager`.

### DIFF
--- a/docs/sources/source-interface.md
+++ b/docs/sources/source-interface.md
@@ -3,26 +3,25 @@
 All source plugins must expose the following:
 
 ## Methods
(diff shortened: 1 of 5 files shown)


fmitchell-r7:
We probably want to update [the documentation on source interfaces](https://github.com/rapid7/propsd/blob/master/docs/sources/source-interface.md) to reflect that `type` is a required property.

fmitchell-r7:
:+1: Thanks for the updated documentation!

PR #39, opened 2016-03-02
Added 'npm run metadata-server' task

dgreene-r7:
Added wrapper around the test-metadata-server so it can be run as an npm task. Added `metadata:host` and `log:level` to the `dev.json` config file.


fmitchell-r7:
Is it worth a node in [the README](https://github.com/rapid7/propsd/blob/master/README.md) letting people know they can run an external EC2 metadata server for testing by executing `npm metadata-server` and that the returned data comes from [a file in the test directory](https://github.com/rapid7/propsd/blob/master/test/data/test-metadata.json)?

dgreene-r7:
The next PR is for an S3 test server and I'm going to create a `Development` section to the README detailing both of these tasks.

fmitchell-r7:
:+1: Looks good to me.

PR #40, opened 2016-03-02
Feature Consul plugin

fmitchell-r7:
This is a first pass at implementing a Consul plugin for propsd. The plugin uses Consul's watch functionality to monitor a local agent for changes.

Services that tag themselves when registering with Consul are grouped. For example, suppose two Elasticsearch nodes are registered with a "production" tag. They have service addresses of 10.0.0.1 and 10.0.0.2. The resulting properties from the plugin will look like

~~~json
{
  "consul": {
    "addresses": ["10.0.0.1", "10.0.0.2"]
  },
  "elasticsearch-production": {
    "addresses": ["10.0.0.1", "10.0.0.2"]
  }
}
~~~


dgreene-r7:
The `Source` interface has been updated (see [source-interface.md](https://github.com/rapid7/propsd/blob/master/docs/sources/source-interface.md)). You should double check what the `Consul` source exposes just to make sure it's consistent.

lib/source/consul.js
```diff
@@ -0,0 +1,292 @@
...
+   * - host (String, default 127.0.0.1): Consul agent's address
+   * - port (Number, default 8500) Consul agent's HTTP(S) port
+   * - secure (Boolean, default: true): enable HTTPS
+   *
+   * @param {Object} options  Options that can be set for the plugin
+   */
+  constructor(options) {
+    super();
+
+    const configuration = options || Object.create(null);
```
dgreene-r7: I've been meaning to ask, because I've seen you use this pattern, but what's the impetus for using `Object.create(null)`?

lib/source/consul.js
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  /**
+   * Return an object describing the Consul plugin's current status
+   *
+   * @return {Object}
+   */
+  status() {
+    return {
+      okay: this._okay,
```
dgreene-r7: To be consistent with the rest of the `Source#status()` output, this should be `ok` rather than `okay`.

lib/source/consul.js
```diff
@@ -0,0 +1,292 @@
...
+      running: this._running
+    };
```
dgreene-r7: There should be an additional `updated` property that represents the last time the underlying data source was updated.

lib/source/consul.js
```diff
@@ -0,0 +1,292 @@
...
+   * @return {Boolean} True if the pooling loop is running
+   * @private
+   */
+  get _running() {
+    return !!this._consul;
+  }
+
+  /* Process updates from Consul's service/list API.
+   *
+   * The data returned Consul looks like this:
```
dgreene-r7: Should probably be: "The data returned **from** Consul looks like this:"

lib/source/consul.js
```diff
@@ -0,0 +1,292 @@
...
+        });
+      } else {
+        watcherInfo.push({
+          name: service,
+          options: {
+            passing: true,
+            service
+          }
+        });
+      }
```
dgreene-r7: This can be simplified because `Array#forEach` won't execute its callback if the array is empty.

```javascript
tags.forEach((tag) => {
  watcherInfo.push({
    name: `${service}-${tag}`,
    options: {
      passing: true,
      service,
      tag
    }
  });
});

if (tags.length === 0) {
  watcherInfo.push({
    name: service,
    options: {
      passing: true,
      service
    }
  });
}
```

lib/source/consul.js
```diff
@@ -0,0 +1,292 @@
...
+    const services = Object.keys(data);
+    const watcherInfo = [];
+
+    services.forEach((service) => {
+      const tags = data[service];
+
+      if (tags.length > 0) {
+        tags.forEach((tag) => {
+          watcherInfo.push({
+            name: service + '-' + tag,
```
dgreene-r7: You can use ES6 string interpolation:
```javascript
name: `${service}-${tag}`
```

Easier to read IMO.

lib/source/consul.js
```diff
@@ -0,0 +1,292 @@
...
+   * [{
+   *   "Node": {
+   *     "Address": "127.0.0.1"
+   *   },
+   *   "Service": {
+   *     "Address": ""
+   *   }
+   * }]
+   *
+   * The Service.Address attribute is provided may or may not be given.
```
dgreene-r7: Not sure what this line means

lib/source/consul.js
```diff
@@ -0,0 +1,292 @@
...
+      }
+      this.properties[name].addresses = addresses.sort();
+    } else {
+      // Empty data means the service has been deregistered.
+      this._unregisterHealthWatcher(name);
+      delete this.properties[name];
+    }
+
+    this._okay = true;
+    this.emit('update', this.properties);
```
dgreene-r7: For consistency, the `update` event should emit `this`.

lib/source/consul.js
```diff
@@ -0,0 +1,292 @@
...
+   *     "Address": ""
+   *   }
+   * }]
+   *
+   * The Service.Address attribute is provided may or may not be given.
+   * If it's provided, it's the address defined by the service. The
+   * Node.Address attribute is the address provided by Consul.
+   *
+   * @param {String} name  The name of the watcher
+   * @param {Object} data  The JSON data from the health/service API
```
dgreene-r7: `length` doesn't exist in `Object.prototype`. My guess is you meant for the `@param` declaration to be `{Array}`, right?

lib/source/consul.js
```diff
@@ -0,0 +1,292 @@
...
+   * @private
+   */
+  _onHealthServiceChange(name, data) {
+    if (data.length > 0) {
+      const addresses = [];
+
+      for (let i = 0; i < data.length; i += 1) {
```
dgreene-r7: Not sure I understand what's going on here. If `data` is an `Object` you can't use `data.length` because it doesn't exist. If it's an array, you could use the more expressive `Array.forEach()`.

lib/source/consul.js
```diff
@@ -0,0 +1,292 @@
...
+        const info = data[i];
+
+        if (info.Service && info.Service.Address) {
+          addresses.push(info.Service.Address);
+        } else if (info.Node && info.Node.Address) {
+          addresses.push(info.Node.Address);
+        }
```
dgreene-r7: What happens if `info` doesn't have `Service` or `Node` keys? Should there be an `else` condition?

lib/source/consul.js
dgreene-r7: As above, for consistency the `update` event should emit `this`.

lib/source/consul.js
```diff
@@ -0,0 +1,292 @@
...
+    }
+  }
+
+  /**
+   * Shut down everything tracking the Consul health/service API.
+   * @private
+   */
+  _shutdownHealthWatchers() {
+    for (const name in this._healthWatchers) {
+      if (this._healthWatchers[name]) {
```
dgreene-r7: I don't think you need this test if you're already iterating over `this._healthWatchers`, especially since calling `_unregisterHealthWatcher()` has the same check.

lib/source/consul.js
fmitchell-r7: I need to figure out the difference between "service unregistered" and "health checks failing". If the service unregisters, we want to delete the watcher. If health checks are failing, we want to keep the watcher around.

lib/source/consul.js
fmitchell-r7: @dgreene-r7 This was updated with your suggestion, but GitHub is still showing the old diff. Not sure why.

dgreene-r7: Looks like the diff was on the close bracket line which never changed. Oh well!

dgreene-r7:
LGTM :+1:

PR #42, opened 2016-03-02
Reorganized test suite

dgreene-r7:
Added an `init` test file with root `before()` and `after()` hooks. This allows us to have a single test metadata server so we don't get transient `EADDRINUSE` errors.

Also added method to stub out `S3.getObject`. This will be deprecated shortly as there are better methods of achieving this by overloading `AWS.S3.prototype.getObject`.

### DIFF
--- a/test/init.js
+++ b/test/init.js
@@ -1,0 +1,19 @@
+/* eslint-env mocha */
+'use strict';
+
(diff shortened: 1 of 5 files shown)


rvanburen-r7:
LGTM :+1:

PR #43, opened 2016-03-02
S3#status() exposes the timer object rather than the interval

dgreene-r7:
Changed the S3#status() output to return the interval rather than the timer. There are edge cases where returning the timer and outputting that data can lead to errors related to nested depth. Plus there's nothing we can actually need from the timer.


dgreene-r7:
From FlowDock:
> Frank 5:23 PM
Reviewed on the train, though my phone's not logged in to GitHub. :shipit:

PR #44, opened 2016-03-02
Feature plugin manager

dgreene-r7:
This is the first of two pull requests for the `PluginManager`. The module is complete, however test coverage only focuses on instantiation and error conditions on startup. The next pull request will focus on testing persistent behavior such as handling events from sources and failures during operation.

### DIFF
--- a/lib/plugin-manager.js
+++ b/lib/plugin-manager.js
@@ -1,0 +1,55 @@
+/* global Log, Config */
+'use strict';
+
+const Metadata = require('./source/metadata');
+//const S3 = require('./source/s3');
+//const Consul = require('./source/consul');
+
+class PluginManager {
+  constructor(storage) {
+    this.storage = storage;
+    this.index = this._registerIndex();
+    this._registerPlugins();
+  }
+
+  _registerIndex() {
+
+  }
+
+  _registerPlugins() {
+    const instance = new Metadata();
+
+    this._registerPluginEvents(instance);
+    this.storage.register(instance);
+    instance.initialize();
+  }
+
+  _registerPluginEvents(plugin) {
+    plugin.on('startup', () => {
+      Log.info(`${plugin.name} started up.`);
+    });
+
+    plugin.on('shutdown', () => {
+      Log.info(`${plugin.name} shut down.`);
+    });
+
+    plugin.on('update', () => {
+      Log.info(`${plugin.name}'s data was updated from its underlying source data.`);
+      this.storage.update();
+    });
+
+    plugin.on('no-update', () => {
+      Log.info(`${plugin.name} has no update to its underlying source data.`);
+    });
+
+    plugin.on('error', (err) => {
+      Log.info(`${plugin.name} encountered the following error: ${err}`);
+    });
+  }
+}
+
+function createPluginManager(sourcePath, storage) {
+  return new PluginManager(sourcePath, storage);
+}
+
+exports.attach = createPluginManager;
--- a/lib/plugin-manager.js
+++ b/lib/plugin-manager.js
@@ -232,6 +232,8 @@ class PluginManager extends EventEmitter {
 
       switch (source.type.toLowerCase()) {
         case 's3':
+
+          // TODO: Support for sources in other buckets
           instance = new S3(Object.assign(source.parameters, {bucket: this.index.bucket}));
           break;
 
--- a/lib/plugin-manager.js
+++ b/lib/plugin-manager.js
@@ -11,7 +11,7 @@ const S3 = require('./source/s3');
 // const Consul = require('./source/consul');
 // const File = require('./source/file');
 
-const DEFAULT_INDEX_INTERVAL = Config.get('index:interval') || 60000;
+const DEFAULT_INDEX_INTERVAL = Config.get('index:interval') || 60000; // eslint-disable-line rapid7/static-magic-numbers
 const DEFAULT_INDEX_BUCKET = Config.get('index:bucket');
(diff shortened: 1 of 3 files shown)


lib/plugin-manager.js
```diff
@@ -0,0 +1,303 @@
...
+'use strict';
+
+const EventEmitter = require('events').EventEmitter;
+const async = require('async');
+const StringTemplate = require('./string-template');
+const Metadata = require('./source/metadata');
+const S3 = require('./source/s3');
+
+// const Consul = require('./source/consul');
+// const File = require('./source/file');
```
fmitchell-r7: Probably worth a "TODO" comment that these are in progress. Or just take out the commented lines.

lib/plugin-manager.js
```diff
@@ -0,0 +1,303 @@
...
+        value = callback(prop, obj[prop]);
+      }
+      collected[prop] = value;
+    }
+  }
+  return collected;
+}
+
+class PluginManager extends EventEmitter {
+  constructor(storage) {
```
fmitchell-r7: Do we want a doc comment here to indicate what type `storage` is?

lib/plugin-manager.js
```diff
@@ -0,0 +1,303 @@
...
+    this._running = false;
+    this._ok = true;
+    this.updateDelay = DEFAULT_CALLBACK_DELAY;
+    this._boundInit = this._init.bind(this);
+  }
+
+  /**
+   * Initialize the PluginManager's index and metadata source
+   */
+  init() {
```
fmitchell-r7: Should we name this `initialize` so it's consistent with plugins?

lib/plugin-manager.js
```diff
@@ -0,0 +1,303 @@
...
+    Log.info('Initializing index and metadata');
+    this._running = true;
+    this._init();
+  }
+
+  /**
+   * Shutdown the plugin manager's sources and then the plugin manager
+   * @param {Boolean} del
```
fmitchell-r7: What does the `del` parameter do?

lib/plugin-manager.js
```diff
@@ -0,0 +1,303 @@
...
+   * Create function that feeds async.parallel (or any async marshalling function)
+   * @param {Metadata|S3} source
+   * @returns {Function}
+   * @private
+   */
+  _createParallelTask(source) {
+    return (c) => {
+      const callBackErr = (err) => {
+        source.removeListener('update', callBackSuccess); // eslint-disable-line no-use-before-define
+        c(err);
```
fmitchell-r7: Can you pass `source.type` into the callback here e.g. `c(err, source.type)`? That might help avoid the "which plugin had an error" question that came up in the `_index` function.

lib/plugin-manager.js
```diff
@@ -0,0 +1,303 @@
...
+      this.storage.register(instance);
+      this._registerSourceEvents(instance);
+      this.emit('source-registered', instance);
+
+      instance.initialize();
+      this.emit('source-initialized', instance);
+    });
+
+    this.index.once('update', () => {
+      // TODO: Implement timeout so that someone spamming updates to the GH repo won't cause a constant rebuild
```
fmitchell-r7: Do you still need this TODO? It looks like you've got a working timeout.

lib/plugin-manager.js
```diff
@@ -0,0 +1,303 @@
...
+        this.storage.update();
+      }, this.updateDelay);
+    });
+
+    source.on('no-update', () => {
+      Log.info(`${source.name} has no update to its underlying source data.`);
+    });
+
+    source.on('error', (err) => {
+      Log.info(`${source.name} encountered the following error: ${err}`);
```
fmitchell-r7: Do we want to log this as `error` instead of `info`?

lib/plugin-manager.js
dgreene-r7: The `del` parameter clears the `Storage` layer's sources array. It was mainly a proposal until we figured out a) how the `PluginManager` shutting down affects its `Storage` object and b) how the `Storage` layer handles clearing its sources.

lib/plugin-manager.js
dgreene-r7: I think adding `source.type` to the callback is a great idea. I'm going to do some investigation to see if it's actually possible to pass additional arguments into the `err` case of `async.parallel`.

lib/plugin-manager.js
```diff
@@ -0,0 +1,302 @@
...
+  /**
+   * Maps instance metadata into the index document
+   * @param {Function} callback
+   * @private
+   */
+  _mapper(callback) {
+    let sources = [];
+
+    try {
+      sources = this.index.properties.sources.map((el) => {
```
jmanero-r7: This isn't going to line up with the behaviors at https://github.com/rapid7/propsd/blob/master/lib/storage.js#L70 and https://github.com/rapid7/propsd/blob/master/lib/source/s3.js#L18 (which isn't parsing the document defined in https://github.com/rapid7/propsd/blob/master/docs/schemas.md#s3-stored-json-configuration-files correctly). `properties` and `sources` should be top-level keys in the document.

dgreene-r7: Can you comment on where the test cases are misconfigured? I copied the test data (in `test/data/s3/`) from the original test data you generated and all tests written against it are passing.

dgreene-r7: I should clarify that the index source isn't being registered with the storage layer by design. It's a special case with a different syntax from the rest of the S3 source documents.

fmitchell-r7:
:+1: Looks good to me.

PR #45, opened 2016-03-03
Default S3 parser wasn't extracting properties object in parsed S3 body

dgreene-r7:
Fixed bug where the default S3 parser was returning the root object parsed from the S3 body instead of the `properties` object within the root object.

We didn't have any test cases built around data that conformed to the schema so this bug was never picked up.

:thumbsup: to @jmanero-r7 for catching this one.


rvanburen-r7:
:+1:

fmitchell-r7:
:+1: Looks good to me.

PR #46, opened 2016-03-03
Updated metadata source status() output

dgreene-r7:
Updated the metadata source so the output from the status() method is consistent with other sources.

### DIFF
--- a/test/metadata.js
+++ b/test/metadata.js
@@ -31,9 +31,7 @@ describe('Metadata source plugin', () => {
 
   it('initializes a timer with the set interval', (done) => {
     this.m.on('update', () => {
-      const status = this.m.status();
-
-      status.interval.should.have.keys(['_called', '_idleNext', '_idlePrev', '_idleStart', '_idleTimeout',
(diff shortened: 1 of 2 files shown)


test/metadata.js
```diff
@@ -31,9 +31,7 @@ describe('Metadata source plugin', () => {
 
   it('initializes a timer with the set interval', (done) => {
     this.m.on('update', () => {
-      const status = this.m.status();
-
-      status.interval.should.have.keys(['_called', '_idleNext', '_idlePrev', '_idleStart', '_idleTimeout',
+      this.m._timer.should.have.keys(['_called', '_idleNext', '_idlePrev', '_idleStart', '_idleTimeout',
```
rvanburen-r7: Shouldn't this be this.interval instead of this.m._timer

dgreene-r7: No, `this.m` is the metadata plugin instance created on line 16. Because I changed the output of `Metadata#status()` to expose the interval rather than the timer itself, we have to dig into "private" properties and test whether the `_timer` object has keys consistent with a timer.

rvanburen-r7:
:+1: 
Thanks for the explanation!

fmitchell-r7:
:+1: Looks good to me.

PR #49, opened 2016-03-03
Added clear() method to each source which clears the source's properties

dgreene-r7:


fmitchell-r7:
If all plugins need to provide `clear` functions, we should update [the documentation](https://github.com/rapid7/propsd/blob/master/docs/sources/source-interface.md) to reflect that.

dgreene-r7:
Good call.

fmitchell-r7:
:+1: Looks good to me.

PR #50, opened 2016-03-03
Fix a timing issue in Consul plugin tests around update time.

fmitchell-r7:
We saw an issue on Travis CI where the test failed.
https://travis-ci.org/rapid7/propsd/builds/113500672

We now ensure that `Consul.status().updated` is set to a time that's later than when the test starts.


dgreene-r7:
:+1:

PR #52, opened 2016-03-04
Integrate core api

dgreene-r7:
Updates to the core API that take into account changes to the storage and plugin manager interfaces.


jmanero-r7:
:+1:

PR #54, opened 2016-03-04
Add a script to fetch and compare properties served from propsd with …

jmanero-r7:
…those served from the to-be-replaced Conqueso role

### DIFF
--- a/test/bin/conqueso-diff
+++ b/test/bin/conqueso-diff
@@ -1,0 +1,83 @@
+#!/usr/bin/env node
+
+/* eslint-disable curly, func-names, no-param-reassign, no-unused-func,
+                  newline-after-var, no-var, one-var, vars-on-top, no-unused-vars,
+                  no-bitwise, no-console, wrap-iife, max-nested-callbacks
+*/
+
+/**
+ * Poll Conqueso and propsd to test for interface inconsistencies
+ */
+var HTTP = require('http');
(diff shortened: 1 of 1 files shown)


test/bin/conqueso-diff
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env node
+
+/* eslint-disable curly, func-names, no-param-reassign, no-unused-func,
+                  newline-after-var, no-var, one-var, vars-on-top, no-unused-vars,
+                  no-bitwise, no-console, wrap-iife, max-nested-callbacks
+*/
```
dgreene-r7: I know it's a test binary but that's a lot of rules to disable.

test/bin/conqueso-diff
```diff
@@ -0,0 +1,83 @@
...
+  var difference = {
+    added: [],
+    removed: []
+  };
+
+  for (var i = 0; i < b.length; i++)
+    if (!~(a.indexOf(b[i]))) difference.added.push(b[i]);
+
+  for (var j = 0; j < a.length; j++)
+    if (!~(b.indexOf(a[j]))) difference.removed.push(a[j]);
```
dgreene-r7: I get what you're doing here but the formatting makes me cry :sob:

dgreene-r7:
Other than formatting sadness it looks good.

PR #55, opened 2016-03-04
Implemented file source

dgreene-r7:


lib/source/file.js
```diff
@@ -0,0 +1,169 @@
...
+
+  initialize() {
+    if (this.service) {
+      return;
+    }
+
+    this._update(this.path);
+
+    this.service = fs.watch(this.path, {persistent: true, recursive: false}, (event, filename) => {
+      this._update(path.resolve(path.dirname(this.path), filename));
```
jmanero-r7: What is this for? `path.dirname(this.path), filename)`

lib/source/file.js
```diff
@@ -0,0 +1,169 @@
...
+    return !!this.service;
+  }
+
+  /**
+   * Updates the properties from the watched file
+   * @param {String} filePath
+   * @private
+   */
+  _update(filePath) {
+    fs.readFile(filePath, (err, data) => {
```
jmanero-r7: The resource's path is already an instance property. There's no reason to pass it as an argument

lib/source/file.js
dgreene-r7: The `filename` arg is just the name of the file and not the complete path. I haven't tested what comes out if you move the file, so my approach may be incomplete.

lib/source/file.js
dgreene-r7: See L82 above. `fs.watch` handles moves as well so I wanted to be able to deal with that in the `fs.watch` callback and pass the path to update to the `_update` method.

lib/source/file.js
jmanero-r7: This is actually discouraged in the docs. Moving the file should not be considered supported behavior.

https://nodejs.org/docs/latest-v0.12.x/api/fs.html#fs_filename_argument

dgreene-r7: Awesome, thanks for finding that. I think we should consider a moved file to be an error case we're not going to handle and simply make sure that `this.path` exists whenever we do an update.

lib/source/file.js
```diff
@@ -0,0 +1,167 @@
...
+  constructor() {
+    this.properties = Object.create(null);
+  }
+
+  /**
+   * Parse the file contents
+   * @param {Buffer} data
+   */
+  update(data) {
+    let properties = {};
```
fmitchell-r7: Might want to use `Object.create(null)` here so results of consistent regardless of if `update` is called or not.

dgreene-r7: Agreed. Good catch.

jmanero-r7: Can you elaborate on this? `Object.create(prototype)` is just setting `__proto__` to the `prototype` argument before returning a new instance of `Object` to you.

jmanero-r7: Bump.

lib/source/file.js
```diff
@@ -0,0 +1,167 @@
...
+      return this;
+    }
+
+    // Release the FSWatcher handle
+    this.service.close();
+
+    delete this.service;
+
+    this._okay = false;
+    this.emit('shutdown');
```
fmitchell-r7: Looks like there's a missing `return this;` e.g. chaining should work if `initialize` has been called.

lib/source/file.js
```diff
@@ -0,0 +1,167 @@
...
+  constructor() {
+    this.properties = Object.create(null);
+  }
+
+  /**
+   * Parse the file contents
+   * @param {Buffer} data
+   */
+  update(data) {
+    let properties = Object.create(null);
```
jmanero-r7: Unhiding my question here:

> fmitchell-r7 added a note 6 minutes ago
> Might want to use Object.create(null) here so results of consistent regardless of if update is called or not.

> @jmanero-r7  Rapid7 member
> jmanero-r7 added a note 2 minutes ago
> Can you elaborate on this? Object.create(prototype) is just setting __proto__ to the prototype argument before returning a new instance of Object to you.

dgreene-r7: `Object.create({})` which is what `properties = {}` does under the hood brings along all of the extra stuff in the `Object` prototype.
```javascript
> const foo = Object.create(null);
> const bar = Object.create({});
> typeof foo.hasOwnProperty;
'undefined'
> typeof bar.hasOwnProperty;
'function'
```

I don't think there's a huge amount of difference creating the object from `null` vs having `Object` in the inheritance chain but who knows. It strikes me as a guard against potential side effects, especially when we don't need anything that the `Object` proto provides.

fmitchell-r7: Yeah, I realized it doesn't actually matter here after I made the comment. We're setting `this.properties` to the properties of `properties` so it's always going to be `undefined` if parsing fails.

dgreene-r7: To me it's just consistency. If line 9 was `this.properties = {};` then line 17 should be too.

PR #56, opened 2016-03-04
Fix PluginManager state machine

fmitchell-r7:
This replaces the bootstrap logic in PluginManager by listening for updates and retrying on a timer. The end result is slightly slower boot time in the worst case. If data from the EC2 metadata or S3 index are unavailable, the PluginManager waits until it gets an update form one or the other before parsing the index for additional sources.

### DIFF
--- a/lib/storage.js
+++ b/lib/storage.js
@@ -76,6 +76,7 @@ class Storage {
    * Remove all source plugins from storage
    */
   clear() {
+    // TODO: WRITE TESTS
     let i = this.sources.length;
 
     while (i--) {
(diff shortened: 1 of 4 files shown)


dgreene-r7:
:+1:

fmitchell-r7:
Waiting until @dgreene-r7 has the plugin cleanup logic worked out before merging this. We want to see how the two interact.

lib/storage.js
```diff
@@ -88,6 +89,10 @@ class Storage {
    * @param {Object} plugin
    */
   unregister(plugin) {
+    // TODO: WRITE TESTS
+    if (typeof plugin.shutdown === typeof (Function)) {
```
jmanero-r7: Use `plugin.shutdown instanceof Function` instead.

jmanero-r7:
:ship: :rocket: :banana:

PR #57, opened 2016-03-04
Fix tests spamming "Server shutting down"

dgreene-r7:
An extra console.log statement snuck in during a merge somewhere. This is what caused tests to spam 'Server shutting down'.


fmitchell-r7:
:+1: Thanks for catching this! :shipit:

PR #58, opened 2016-03-04
Fix tests spamming "Server shutting down" for real

fmitchell-r7:
#57 accidentally deleted the extra log statement that was outside the loop. This removes the one inside the loop. So now there's only one "Server shutting down" statement.


dgreene-r7:
:+1:

PR #59, opened 2016-03-04
Config properties

dgreene-r7:
Merge properties delivered via the Config object with the metadata properties.

We'll eventually phase this out in favor of using a `File` source, but it's a good stop-gap until that's fully implemented and being polled in the `PluginManager`.


fmitchell-r7:
Can you drop a note in the README describing the use case for this?

jmanero-r7:
:ship:

fmitchell-r7:
:+1: Looks good to me.

PR #60, opened 2016-03-04
Adding coveralls.io support

dgreene-r7:
Added coveralls and mocha clove reporter. Added report-coverage npm task that runs coverage reports and sends it to coveralls.io. Added that task to Travis if linting and tests succeed.


fmitchell-r7:
:+1: Looks good to me.

Issue #61, opened 2016-03-07
Conqueso API needs to format properties from Consul plugin


fmitchell-r7:
The properties [returned by the Consul plugin](https://github.com/rapid7/propsd/blob/2154e00472f33e449a128232b6075edcce97dc54/lib/source/consul.js#L201-L215) have the format "service.addresses" or "service-tag.addresses" where "service" is the name of the service as it appears in Consul, "-tag" is an optional tag if the service has one, and "addresses" is an array of IP addresses.

To maintain compatibility with existing Conqueso clients, the [Conqueso API should reformat properties](https://github.com/rapid7/propsd/blob/2154e00472f33e449a128232b6075edcce97dc54/lib/control/v1/conqueso.js#L29-L37) returned by the Consul plugin as "conqueso.name.ips=addresses" where "name" is either the service name if the service isn't tagged, or the tag if the service has tags, and "addresses" is a comma separated list of IP addresses.

PR #62, opened 2016-03-07
Feature: Reusable Consul stub

fmitchell-r7:
This is preliminary work for #61. The Consul stub has been refactored so it's usable outside the Consul tests. This will enable the Conqueso tests to mock data from the Consul plugin and ensure it's formatted correctly. The sub has also been documented at the API level.


test/utils/consul-stub.js
```diff
@@ -0,0 +1,127 @@
+'use strict';
```
rvanburen-r7: Quick question, shouldn't we have `/* eslint-env mocha */

fmitchell-r7: I don't think it's necessary, no. As far as I know, mocha won't run up anything in the "./utils" folder by default, and this file shouldn't be using anything mocha specific.

rvanburen-r7: Ah, thanks for the explanation!

rvanburen-r7:
LGTM, but you may still want another pair of :eyes: on this.

dgreene-r7:
Very cool. Nice job!

PR #63, opened 2016-03-07
Debug task

dgreene-r7:
This PR adds a `debug-server` npm task that spins up test metadata and test S3 servers and runs an instance of propsd using a dev config file.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const nconf = require('nconf');
-const Path = require('path');
 
 const defaults = {
   index: {
@@ -28,9 +27,12 @@ const defaults = {
 };
 
 function loadConfigs(path) {
-  const configs = nconf.file(path);
+  let configs = nconf.defaults(defaults);
 
-  configs.defaults(defaults);
+  if (path) {
+    configs = nconf.file(path);
+  }
+
   return configs;
 }
 
--- a/lib/config.js
+++ b/lib/config.js
@@ -28,8 +28,7 @@ const defaults = {
 };
 
 function loadConfigs(path) {
-  const configPath = path || Path.resolve(__dirname, '../config/dev.json');
-  const configs = nconf.file(configPath);
+  const configs = nconf.file(path);
 
   configs.defaults(defaults);
   return configs;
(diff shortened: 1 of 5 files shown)


rvanburen-r7:
LGTM, but you should get another set of :eyes: on it besides mine.

lib/config.js
```diff
@@ -28,10 +27,12 @@ const defaults = {
...
 function loadConfigs(path) {
-  const configPath = path || Path.resolve(__dirname, '../config/dev.json');
-  const configs = nconf.file(configPath);
+  let configs = nconf.defaults(defaults);
+
+  if (path) {
+    configs = nconf.file(path);
+  }
 
-  configs.defaults(defaults);
```
fmitchell-r7: Looks like we lost the behavior where defaults got filled in for config bits that weren't in the file. Was that intentional?

test/bin/s3-server.js
```diff
@@ -73,55 +83,101 @@ function createTempDirForBucket() {
...
 
+/**
+ * Handles cleanup on app exit
+ */
 function exitHandler() {
-  console.log('Cleaning up temp directory'); // eslint-disable-line no-console
+  Log.info('Cleaning up temp directory');
   cleanupTempDir();
   process.exit(0);
 }
```
fmitchell-r7: Do we need API docs for the `createBucket` function so it matches up with the `emptyBucket` function?

lib/config.js
dgreene-r7: See line 30

dgreene-r7: Good catch. I moved it to line 30 but forgot to change variable names.

fmitchell-r7:
Thanks for adding this! Is it worth putting a note in the README [like we have for the S3 server](https://github.com/rapid7/propsd#s3-server) so people know this is available for local integration testing?

lib/config.js
```diff
@@ -28,10 +27,12 @@ const defaults = {
 };
 
 function loadConfigs(path) {
-  const configPath = path || Path.resolve(__dirname, '../config/dev.json');
-  const configs = nconf.file(configPath);
+  let configs = nconf.defaults(defaults);
```
jmanero-r7: This is a weird usage of nconf. Generally, configuration is a singleton, so the nconf package actually exports an instantiated config object, which you attach sources to:

```javascript

var Config = globals.Confog = require('nconf');
Config.argv(...);
Config.file(...);
Config.defaults({
...
});

module.exports = Config;
```

dgreene-r7:
I'm going to close this PR, fix the config service, and then reopen.

PR #64, opened 2016-03-07
Fix nvmrc version

dgreene-r7:
At some point our nvmrc file drifted from the most current version of the 4.x branch so this sets it back.


fmitchell-r7:
:+1: Yay! Thanks for locking this down. :shipit:

PR #65, opened 2016-03-07
Deprecate config service

dgreene-r7:
This PR addresses #67 by removing the config service and tests around it and replacing it with a few lines in `server.js`, `init.js`, and `metadata-server.js` and a `defaults.json` file in the `config` directory.


fmitchell-r7:
Probably worth a note in the description for this pull request about _why_ we're deprecating / removing config.js. And do we need to update the README, or is the behavior still the same?

dgreene-r7:
Behavior is still the same but the reason we're deprecating should definitely be captured. Suggestions about where?

fmitchell-r7:
@dgreene-r7 You can open an issue on it to capture the why. Then update the description for this pull request to note that you're fixing that issue. That provides a place for conversation (the issue) and creates links between the two.

fmitchell-r7:

PR #66, opened 2016-03-08
Fix: Namespaced Consul properties

fmitchell-r7:
As part of #61, the Conqueso API needs to be able to recognize that properties it's formatting are from Consul. This fixes the Consul plugin so it namespaces its generated properties by enclosing them in a "consul" object.

I've kind of viewing this as a temporary fix for now. Long term, we may want to come up with a design around an internal "properties" structure. That would allow us to define rendering rules based on property metadata.

### DIFF
--- a/lib/source/consul.js
+++ b/lib/source/consul.js
@@ -21,6 +21,7 @@ class Consul extends EventEmitter {
     this.type = 'consul';
     this.name = 'consul';
     this.properties = Object.create(null);
+    this.properties.consul = Object.create(null);
     this._okay = false;
     this._updated = null;
   }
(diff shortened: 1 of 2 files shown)


dgreene-r7:
In my mind, this sort of thing is better handled with a `ConsulParser` that takes `Consul#properties` and does any transformation or formatting.

fmitchell-r7:
@dgreene-r7 Any objection to my opening a new issue around the refactoring and us shipping this as is?

dgreene-r7:
Up to you. I think you should capture it in an issue, regardless.

fmitchell-r7:
I added #69 to capture the refactor work.

dgreene-r7:
:shipit:

PR #68, opened 2016-03-08
Fixed bug where path resolution for the config file wasn't working correctly

dgreene-r7:
Ref #65


fmitchell-r7:
:+1: Thanks for catching this! :shipit:

PR #71, opened 2016-03-08
Feature dev server task

dgreene-r7:
This PR creates self-contained S3 and Metadata servers to mock dependencies and allow local development. Simply run `npm run dev-server` and you should be all set.

### DIFF
--- a/test/bin/s3-server.js
+++ b/test/bin/s3-server.js
@@ -91,7 +91,7 @@ function copyFiles(bucket) {
   const walker = walk.walk(path);
 
   walker.on('file', (root, fileStats, next) => {
-    const pathToFile = `${Path.relative(path, root)}/${fileStats.name}`;
+    const pathToFile = Path.join(Path.relative(path, root), fileStats.name);
     const stream = fs.createReadStream(Path.join(path, pathToFile));
 
     awsClient.putObject({
@@ -128,6 +128,7 @@ function emptyBucket(bucket) {
 
 /**
  * Handles cleanup on app exit
+ * @param {Error} err
  */
 function exitHandler(err) {
   let code = 0;
--- a/test/bin/s3-server.js
+++ b/test/bin/s3-server.js
@@ -141,6 +141,11 @@ function exitHandler(err) {
   process.exit(code);
 }
 
+function onFileChange(bucket) {
+  emptyBucket(bucket);
+  copyFiles(bucket);
+}
+
 function createBucket(bucket) {
   awsClient.createBucket({Bucket: bucket}, (createBucketErr) => {
     if (createBucketErr) {
@@ -148,10 +153,9 @@ function createBucket(bucket) {
     }
     copyFiles(bucket);
 
-    chokidar.watch(path, {persistent: true}).on('change', () => {
-      emptyBucket(bucket);
-      copyFiles(bucket);
-    });
+    // Watch for both change and unlink events.
+    chokidar.watch(path, {persistent: true}).on('change', () => onFileChange(bucket));
+    chokidar.watch(path, {persistent: true}).on('unlink', () => onFileChange(bucket));
   });
 }
 
--- a/test/bin/s3-server.js
+++ b/test/bin/s3-server.js
@@ -137,7 +137,9 @@ function exitHandler(err) {
     Log.error(err, err.stack);
     code = 1;
   }
-  Log.info('Cleaning up temp directory');
+  const tmpDirBucketPath = Path.resolve(os.tmpdir(), `propsd-temp-s3-server`);
+
+  Log.info(`Cleaning up temp directory: ${tmpDirBucketPath}`);
   cleanupTempDir();
   process.exit(code);
 }
--- a/test/bin/s3-server.js
+++ b/test/bin/s3-server.js
@@ -129,10 +129,16 @@ function emptyBucket(bucket) {
 /**
  * Handles cleanup on app exit
  */
-function exitHandler() {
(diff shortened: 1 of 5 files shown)


test/bin/s3-server.js
```diff
@@ -73,55 +83,107 @@ function createTempDirForBucket() {
...
+/**
+ * Copies files from the watched folder the the "bucket"
+ * @param {String} bucket
+ */
+function copyFiles(bucket) {
+  const walker = walk.walk(path);
+
+  walker.on('file', (root, fileStats, next) => {
+    const pathToFile = `${Path.relative(path, root)}/${fileStats.name}`;
+    const stream = fs.createReadStream(`${path}/${pathToFile}`);
```
fmitchell-r7: Any reason not to use `Path.join` instead of string interpolation?

test/bin/s3-server.js
```diff
@@ -73,55 +83,107 @@ function createTempDirForBucket() {
...
+    });
   });
+}
 
-  client.run((serverErr, host, p) => {
-    if (serverErr) {
+function init() {
+  fs.stat(path, (err, stats) => {
+    if (err || !stats.isDirectory()) {
       process.exit(1);
```
fmitchell-r7: Does it make sense to log something here before exiting so the user knows why the process is exiting?

test/bin/s3-server.js
```diff
@@ -73,55 +83,107 @@ function createTempDirForBucket() {
...
+      silent: false,
+      directory: tmpDir
+    });
 
-    awsClient.createBucket({Bucket: bucket}, (createBucketErr) => {
-      if (createBucketErr) {
-        throw createBucketErr;
+    client.run((serverErr, host, p) => {
+      if (serverErr) {
+        process.exit(1);
```
fmitchell-r7: Does it make sense to log something here before exiting so the user knows why the server is exiting?

test/bin/s3-server.js
dgreene-r7: Nope, not in particular. Brainfarted and forgot that it existed.

test/bin/s3-server.js
```diff
@@ -73,55 +83,113 @@ function createTempDirForBucket() {
...
-    process.exit(1);
+/**
+ * Copies files from the watched folder the the "bucket"
+ * @param {String} bucket
+ */
+function copyFiles(bucket) {
+  const walker = walk.walk(path);
+
+  walker.on('file', (root, fileStats, next) => {
+    const pathToFile = `${Path.relative(path, root)}/${fileStats.name}`;
```
fmitchell-r7: Can you use `Path.join` here as well?

test/bin/s3-server.js
```diff
@@ -73,55 +83,113 @@ function createTempDirForBucket() {
...
+    chokidar.watch(path, {persistent: true}).on('unlink', () => onFileChange(bucket));
   });
+}
 
-  client.run((serverErr, host, p) => {
-    if (serverErr) {
+function init() {
+  fs.stat(path, (err, stats) => {
+    if (err || !stats.isDirectory()) {
+      Log.error('Path is not a directory');
```
fmitchell-r7: Do we want to log the path as part of the error message so the user knows which one is invalid?

test/bin/s3-server.js
```diff
@@ -73,55 +83,113 @@ function createTempDirForBucket() {
...
+    const pathToFile = `${Path.relative(path, root)}/${fileStats.name}`;
+    const stream = fs.createReadStream(Path.join(path, pathToFile));
+
+    awsClient.putObject({
+      Bucket: bucket,
+      Key: pathToFile,
+      Body: stream
+    }, (err) => {
+      if (err) {
+        Log.info(err, err.stack);
```
fmitchell-r7: Should we log this at the `error` level instead of `info`?

test/bin/s3-server.js
```diff
@@ -73,55 +83,113 @@ function createTempDirForBucket() {
...
+      silent: false,
+      directory: tmpDir
+    });
 
-    awsClient.createBucket({Bucket: bucket}, (createBucketErr) => {
-      if (createBucketErr) {
-        throw createBucketErr;
+    client.run((serverErr, host, p) => {
+      if (serverErr) {
+        Log.error(serverErr);
```
fmitchell-r7: Do we want to log `serverErr.stack` as well? Would that be useful?

test/bin/s3-server.js
```diff
@@ -73,55 +83,113 @@ function createTempDirForBucket() {
...
+}
+
+/**
+ * Handles cleanup on app exit
+ */
+function exitHandler(err) {
+  let code = 0;
+
+  if (err) {
+    Log.error(err);
```
fmitchell-r7: Do we want to log `err.stack` as well? Would that be useful?

test/bin/s3-server.js
```diff
@@ -73,55 +83,113 @@ function createTempDirForBucket() {
...
   }
-  const bucket = getBucketName(path);
-  const tmpDir = createTempDirForBucket();
-
-  const client = new S3rver({
-    port,
-    hostname,
-    silent: false,
-    directory: tmpDir
+  Log.info('Cleaning up temp directory');
```
fmitchell-r7: Do we have the temp directory name so we can log that here too? Might be useful for the user to know what directory they need to manually clean up if the clean up process fails.

fmitchell-r7:
:+1: Thanks for getting all these development bits working together. :shipit:

PR #73, opened 2016-03-08
Fix: Better Conqueso compatibility

fmitchell-r7:
Another piece of puzzle around #61 to get the Conqueso API formatting output correctly. This translates from IP addresses as they're returned in the Consul plugin into the format defined by Conqueso. Currently on non-tagged services are supported. Tagged services will be handled in a separate pull request. There's a pending test to make sure I don't forget to do that work.

### DIFF
--- a/lib/control/v1/conqueso.js
+++ b/lib/control/v1/conqueso.js
@@ -6,7 +6,7 @@ const HTTP_OK = 200;
 const HTTP_METHOD_NOT_ALLOWED = 405;
 
 /**
- * Format the given data as Java properites
+ * Format the given data as Java properties
  *
  * @param {Object} data
  * @return {String}
--- a/lib/control/v1/conqueso.js
+++ b/lib/control/v1/conqueso.js
@@ -36,25 +36,19 @@ function makeJavaProperties(data) {
  *   "conqueso.service.ips": "x.x.x.x, y.y.y.y"
  * }
  *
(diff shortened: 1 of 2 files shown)


dgreene-r7:
Looks good :+1:

lib/control/v1/conqueso.js
```diff
@@ -6,6 +6,74 @@ const HTTP_OK = 200;
...
+ *
+ * @param {Object} properties  Flattened properties as returned from Storage#properties
+ * @return {Object}             Properties with Consul address converted to Conqueso ones
+ */
+function transformAddresses(properties) {
+  const results = {};
+
+  for (const key in properties) {
+    if (properties.hasOwnProperty(key)) {
+      if (key.startsWith('consul') && key.endsWith('addresses')) {
```
jmanero-r7: This could be performed on the deep properties object instead of doing this with string-groking after the properties have been flattened:

```javascript
var consulServices = {}
object.keys(properties.consul).forEach(function(service) {
 consulServices[['conqueso', service, 'ips'].join('.')] = properties.consul[service].join(',');
});

// Merge consulServices into 
```

fmitchell-r7: Thanks! That's what I wanted to do originally but couldn't figure out how. I refactored the code so Conqueso formatted addresses are not injected into the returned Java properties list based on the list of Consul services.

jmanero-r7:
:+1:

PR #74, opened 2016-03-08
PD-6304 Add Rakefile to release/package props.d

athompson-r7:
This pull requests aims to setup a "oneliner" that can be placed on the build slaves so that we can build packages of props.d.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,0 +1,89 @@
+require 'json'
+require 'fileutils'
+require 'mkmf'
+require 'aws-sdk'
+require 'logger'
+
+include FileUtils
+
+task :version do
+ @version = JSON.parse(File.read('package.json'))['version']
+end
+
+task :base_dir do
+  @base_dir = File.dirname(File.expand_path(__FILE__))
+end
+
+task :install do
+  sh 'npm install --only=prod'
+end
+
+task :shrinkwrap => [:install] do
+  sh 'npm shrinkwrap'
+end
+
+task :pack => [:shrinkwrap] do
+  sh 'npm pack'
+end
+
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -0,0 +1,89 @@
...
+
+task :shrinkwrap => [:install] do
+  sh 'npm shrinkwrap'
+end
+
+task :pack => [:shrinkwrap] do
+  sh 'npm pack'
+end
+
+desc 'Release props.d and prepare to create a release on github.com'
```
dgreene-r7: Seeing `props.d` reminds me that we should go through and standardize the naming across the entire project. I'm going to open an issue to address this, but in the meantime, can you change it to `propsd`?

dgreene-r7:
You've got hard-coded project-related strings throughout this file that might be better suited as a task that reads from `package.json`. I'm thinking specifically `props.d` vs `propsd` and how we can centralize everything around the canonical project name.

dgreene-r7:
:+1:

Rakefile
```diff
@@ -0,0 +1,100 @@
+require 'json'
+require 'fileutils'
+require 'mkmf'
```
fmitchell-r7: Are we using `mkmf` anywhere? I'm not seeing any references, but it might be pulling it helper stuff.

athompson-r7: It provides `find_executable`; we are using `find_executable` at line 59.

fmitchell-r7:
:+1: Thanks for getting this all working! :shipit:

athompson-r7:
Thanks guys!

PR #76, opened 2016-03-09
Fix: Handle tagged Conqueso IPs

### LINKED ISSUE #61: Conqueso API needs to format properties from Consul plugin

fmitchell-r7:
This fixes #61. Services registered with Consul that are tagged with their cluster name are now exposed as `conqueso.cluster.ips=x.x.x.x` on the `/v1/conqueso` end point. Services that don't have tags are exposed with the service name as the cluster name.

### DIFF
--- a/lib/control/v1/conqueso.js
+++ b/lib/control/v1/conqueso.js
@@ -44,7 +44,9 @@ function injectConquesoAddresses(properties) {
 
   if (properties.consul) {
     Object.keys(properties.consul).forEach((service) => {
-      conquesoIps[`conqueso.${service}.ips`] = properties.consul[service].addresses.join(',');
+      const cluster = properties.consul[service].cluster;
+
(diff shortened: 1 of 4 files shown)


dgreene-r7:
:+1:

PR #77, opened 2016-03-09
Hook up Consul source plugin

dgreene-r7:
This PR registers the Consul plugin. Right now we're registering it outside of the index loop so it will always be registered, regardless of the contents of the index.


fmitchell-r7:
:+1: Looks good to me.

PR #84, opened 2016-03-10
Consul source should not be shutdown on index update

### LINKED ISSUE #79: Consul plugin should not be shutdown when S3 index changes
Because the Consul plugin is registered with Storage as part of the PluginManager, when the index document in S3 changes, the Consul plugin gets shut down. It's then never started back up again. We might want to move the Consul plugin into the index document and have a check in the PluginManager to make sure there's only one Consul source configured.

dgreene-r7:
This resolves #79. The tradeoff is that the consul source needs to be enumerated in the index document in the correct location. The `PluginManager` will pick up the first occurrence of the source plugin and ignore all others so make sure that the consul plugin is positioned correctly in the index so it doesn't get overridden.

### DIFF
--- a/test/plugin-manager.js
+++ b/test/plugin-manager.js
@@ -212,7 +212,7 @@ describe('Plugin manager', function () {
     manager.initialize();
   });
 
-  before(() => {
+  it('only allows one instance of the consul source', (done) => {
     const indexWithMultipleConsulSources = JSON.parse(fakeIndexResponse.Body.toString());
 
     // Add a bunch of consul sources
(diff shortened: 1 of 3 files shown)


test/plugin-manager.js
```diff
@@ -204,6 +212,43 @@ describe('Plugin manager', function () {
     manager.initialize();
   });
 
+  before(() => {
```
fmitchell-r7: Do we need/want to scope this in its own `describe` block so the `before` and `after` functions don't interact with other tests?

jmanero-r7:
:+1:

PR #85, opened 2016-03-10
Fix: PluginManager TODO tests

fmitchell-r7:
This adds more tests around the PluginManager. We make sure old plugins are shut down when the index changes. We also make sure plugins are added and removed when the index changes.


dgreene-r7:
Looks good!

PR #86, opened 2016-03-10
Fix: Can merge Consul plugin properties

### LINKED ISSUE #80: Consul plugin's properties are missing hasOwnProperty function
The Consul plugin's properties don't have a `hasOwnProperty` function. This means they don't get merged when Storage is updated. The Consul plugin's properties object should be changed to have `Object` as its prototype and a test added to make sure that works.

fmitchell-r7:
This fixes #80. Usage of `Object.create(null)` has been replaced with `{}` in the Consul plugin so the `hasOwnProperty` checks in the storage engine work correctly.

### DIFF
--- a/lib/source/consul.js
+++ b/lib/source/consul.js
@@ -91,7 +91,7 @@ class Consul extends EventEmitter {
    * @param {Object} options  Options that can be set for the plugin
    */
   configure(options) {
-    const params = options || Object.create(null);
+    const params = options || {};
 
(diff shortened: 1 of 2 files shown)


dgreene-r7:
:+1: looking good

jmanero-r7:
:shipit:

dgreene-r7:
I'm going to merge this for @fmitchell-r7.

Issue #82, opened 2016-03-10
Consul plugin's logged update messages aren't helpful


fmitchell-r7:
When the Consul plugin updates, the following message is written to the log
~~~text
info: consul's data was updated from its underlying source data.
~~~
This isn't particularly helpful, since it doesn't tell you what was updated. The message should be changed to log which Consul watcher triggered the update. We might want to have this be a level message instead of an info level as well.

PR #87, opened 2016-03-10
Fix logging levels

dgreene-r7:
Addresses #82. Changed some logging levels to be more significant. Operations (startup, shutdown, update) should be logged as `info` from their respective source. Source status information emitted from the `PluginManager` are logged as `verbose` to maintain some separation from `debug` level logging which is fired during source polling.

I also updated the consul source so that it's logging is a little bit more meaningful.


jmanero-r7:
:+1:

PR #91, opened 2016-03-11
Fix: Remove translated Consul properties

### LINKED ISSUE #90: Remove translated Consul plugin properties from Conqueso output
The Conqueso API [translates Consul plugin output](https://github.com/rapid7/propsd/blob/master/lib/control/v1/conqueso.js#L26-L54) into Conqueso compatible output. We should remove the `consul.` output that was translated so it doesn't show up in the final properties set. This will avoid exposing Conqueso clients to properties they don't care about.

fmitchell-r7:
This fixes #90. Consul properties that have been translated into the Conqueso compatibility layer are removed from the formatted output.

### DIFF
--- a/lib/control/v1/conqueso.js
+++ b/lib/control/v1/conqueso.js
@@ -6,6 +6,39 @@ const HTTP_OK = 200;
 const HTTP_METHOD_NOT_ALLOWED = 405;
 
 /**
+ * Deep copy an object
+ * https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm
+ *
+ * @param {Object} object  The object to be copied
+ * @return {Object}        A deep copy of the provided object
+ */
+function clone(object) {
+  if (!(object instanceof Object)) {
+    return object;
+  }
+
+  const Constructor = object.constructor;
+  let clonedObject = null;
+
+  switch (Constructor) {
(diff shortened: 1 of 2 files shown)


dgreene-r7:
:+1:

jmanero-r7:
Please hold

lib/control/v1/conqueso.js
```diff
@@ -48,6 +49,7 @@ function injectConquesoAddresses(properties) {
 
       conquesoIps[`conqueso.${cluster}.ips`] = properties.consul[service].addresses.join(',');
     });
+    delete properties.consul; // eslint-disable-line no-param-reassign
```
jmanero-r7: It looks like `properties` a reference to the object managed by the Storage module. Mutating it here is likely to have strange effects on other consumers that expect it's `consul` key to not disappear whenever this handler is executed.

We should do something like this insterad:

```
// Construct a new Object with only 'conqueso.SERVICE.ips=list,of,addresses` key/values
const conquesoIps = translateConquesoAddresses(properties);

const conquesoProperties = Object.assign({}, properties, conquesoIps);
delete conquesoProperties.consul;

makeJavaProperties(flatten(conquesoProperties));
...
```

jmanero-r7:
Looks good :+1:

PR #92, opened 2016-03-11
Feature: Package default config file

### LINKED ISSUE #81: Default configuration file is missing from Debian package
When we package propsd as a Debian package, we're not getting [the default configuration file](https://github.com/rapid7/propsd/blob/9aaed11e2ff0440e1619e82c1a1cfc7406fffe1a/config/defaults.json) included. That file should go in a "config" folder adjacent to the "bin" folder once the package is installed.

fmitchell-r7:
This fixes #81. The "defaults.json" config file is now being included in the Debian package. In addition, temporary files in the "pkg/" folder won't be accidentally dumped into source control.


jmanero-r7:
:+1:

dgreene-r7:
:shipit:

PR #93, opened 2016-03-11
Fix ordering ips in conqueso diff

dgreene-r7:
This PR resolves an issue in the `conqueso-diff` tool where values that were lists (IPs, etc.) weren't being sorted and were showing as false positives.

### DIFF
--- a/test/bin/conqueso-diff
+++ b/test/bin/conqueso-diff
@@ -7,26 +7,28 @@
 const HTTP = require('http');
 const ONE_MINUTE = 60000;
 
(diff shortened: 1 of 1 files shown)


fmitchell-r7:
What's the goal behind the refactor to use `Promise` objects?

dgreene-r7:
The previous method of doing things had nested callbacks that eslint didn't like. It wasn't a big deal but I felt like doing it with a `Promise` made it a little cleaner.

fmitchell-r7:
:+1:

PR #94, opened 2016-03-11
Feature: debian package should recommend node

### LINKED ISSUE #83: Debian package should recommend Node.js instead of installing the binary
The Debian package installs a binary for Node.js. This means if you create a Debian package on OS X, you can't use the package on Linux. We should change the package to not install a binary, but just recommend that Node.js be installed.

According to the [Node.js instructions for installing a Debian package](https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions) the package name is "nodejs". We should be able to update [the fpm command](https://github.com/rapid7/propsd/blob/52587f49a909c320b39368c156e5cd652af648fa/Rakefile#L76) with `--deb-recommends nodejs` and remove the `node` binary.

dgreene-r7:
This resolves #83.

Instead of packing the `node` binary with the deb, we should recommend that Node be installed.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -22,10 +22,6 @@ def install_dir
   ::File.join('pkg', 'opt', name)
 end
 
-def bin_dir
-  ::File.join('pkg', 'usr', 'bin')
-end
-
 def config_dir
   ::File.join(install_dir, 'config')
 end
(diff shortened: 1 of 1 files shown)


fmitchell-r7:

athompson-r7:
The version of nodejs in the ubuntu repos is way behind.  If you want to move this to recommends you should at least lock the version.

PR #95, opened 2016-03-14
Fix: conqueso diff property parsing

dgreene-r7:
This PR applies some transformation and sorting around properties so that we don't see false positives in the diff output.

Specifically, it sorts IP addresses (conqueso doesn't sort by default) and handles JavaScript trying to be helpful when it truncates `75.0.toString()` to `'75'`.

Archaius maps properties to Java types and reads all values from the conqueso/propsd endpoint as strings anyway, so the diffs produced by this tool truncate values with trailing zeroes that are produced by conqueso in the interest of matching values correctly.

### DIFF
--- a/test/bin/conqueso-diff
+++ b/test/bin/conqueso-diff
@@ -75,8 +75,8 @@ Array.prototype.truncateNumericProperties = function truncateNumericProperties()
 /* eslint-enable no-extend-native */
 
 function compareProperties(a, b) {
-  a = sortDelimitedValues(a.sort()); // eslint-disable-line no-param-reassign
-  b = sortDelimitedValues(b.sort()); // eslint-disable-line no-param-reassign
+  a = a.sort().sortDelimitedValues().truncateNumericProperties(); // eslint-disable-line no-param-reassign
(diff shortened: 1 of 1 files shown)


test/bin/conqueso-diff
```diff
@@ -33,26 +33,50 @@ function requestProperties(uri) {
...
+Array.prototype.sortDelimitedValues = function sortDelimitedValues() {
+  return apply(this, (val) => val.split(',').length > 1, (val) => val.split(',').sort());
+};
+
+/**
+ * Truncates numeric properties for conqueso property string float -> js float comparison
+ * @returns {Array}
+ */
+Array.prototype.truncateNumericProperties = function truncateNumericProperties() {
+  return apply(this, (val) => !isNaN(val) && isFinite(val), (val) => parseFloat(val));
```
fmitchell-r7: I'm pretty sure you can write this with `reduce`.
~~~javascript
return this.reduce((memo, val) => {
  if (!isNaN(val) && isFinite(val) {
    memo.push(parseFloat(val);
  } else {
    memo.push(val);
  }
  return memo;
});
~~~

fmitchell-r7:

PR #96, opened 2016-03-14
Fix: version constraint on deb recommends

dgreene-r7:
Updated the `rake deb` task to include `depends` rather than `recommends` as well as other metadata like `license`, `url`, and `description`.

Node version dependency is pulled from the project `.nvmrc` file.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -89,7 +89,7 @@ desc 'Cleanup release and package artifacts'
 task :clean do
   rm_f 'npm-shrinkwrap.json'
   rm_f Dir["#{name}-*.tgz"]
-  rm_rf 'pkg' 
+  rm_rf 'pkg'
 end
 
 task :default => [:clean, :release, :package]
--- a/Rakefile
+++ b/Rakefile
@@ -34,6 +34,10 @@ def target_version
   ::File.read(::File.join(@base_dir, '.nvmrc')).strip()
 end
 
+def max_version
+   target_version.split('.').first.to_f + 1
+end
+
 def install_dir
   ::File.join('pkg', 'opt', name)
 end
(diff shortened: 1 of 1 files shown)


dgreene-r7:
We should probably figure out the `vendor` and `maintainer` fields as well.

Rakefile
```diff
@@ -68,7 +84,20 @@ task :chdir_pkg => [:package_dirs] do
 end
 
 task :deb => [:chdir_pkg, :propsd_source] do
-  sh "fpm --deb-no-default-config-files --deb-recommends nodejs -s dir -t deb -n \"#{name}\" -v #{version} opt/"
+  command = [
+    'fpm',
+    '--deb-no-default-config-files',
+    "--depends \"nodejs = #{target_version}\"",
```
fmitchell-r7: Is there a way to make this fuzzy? Right now, our target version is 4.3.2, but I'm pretty sure stuff would work on 4.4.0 as well. Would making this be `>= 4.x.x` do the right thing?

athompson-r7:
LGTM

Rakefile
dgreene-r7: There is *ish* by doing `--depends "nodejs >= 4" --depends "nodejs << 5"` but I have two concerns:
1. `dpkg` doesn't respect semver afaik so saying `>=4` might just pick up `4.0.0` or `4.0.x` or `4.x.x` depending on what the ppa is serving.
1. Pegging the depends version to our development version at least guarantees that we're deploying with what we've tested against.

athompson-r7: I would take a look at this:
https://debian-handbook.info/browse/stable/sect.package-meta-information.html

For example 1.1.1 is not equal to 1.1.1-ubuntu~patch10.303

dgreene-r7: That doesn't explain whether `>= 4` includes `4.x.x`, `4.x`, or only 4+.

fmitchell-r7: Based on the docs referenced above, I think we can do semverish pinning with `--depends 'nodejs >= 4.4, nodejs << 5.0'`.

fmitchell-r7:
:+1: Thanks for working through the wackiness that is Debian dependency resolution! :shipit:

PR #99, opened 2016-03-14
Feature: request logging

### LINKED ISSUE #98: Request logging
It would be helpful to have request logging so we can see what endpoints are accessed and how frequently. Eventually when we do service-based endpoints it'll be helpful to know how often different services hit endpoints.

dgreene-r7:
Using the [morgan](https://github.com/expressjs/morgan) lib to do request logging. Refactored the logging service to be more generic and less dependent on `nconf`-like objects.

This PR resolves #98.


bin/server.js
```diff
@@ -16,39 +16,46 @@ const args = require('yargs')
...
-const server = http.createServer(app);
-const storage = new (require('../lib/storage'))();
-const manager = new PluginManager(storage);
+// Add request logging middleware
+const accessLog = Logger.attach(global.Config.get('log:access:level'), global.Config.get('log:access:filename'));
 
-manager.on('error', (err) => {
-  Log.error(err);
-});
+app.use(Logger.logRequests((message) => accessLog.verbose(message)));
```
fmitchell-r7: What's the `.verbose` method doing? Does that mean messages are only logged at the verbose level?

dgreene-r7: That's right. `accessLog` is just an instance of `Winston`.

We can change it to log at `debug` if you think `verbose` is too high. See [Winston logging levels](https://github.com/winstonjs/winston#logging-levels).

fmitchell-r7: I guess access logs are `verbose` by design, so that makes sense.

dgreene-r7: Alternativel we could do something like

```javascript
app.use(Logger.logRequests((message) => accessLog.log(global.Config.get('log:access:level'), message)));
```

This way we'll log everything to the access log at the level we're defining in the config file.

fmitchell-r7: :+1: Yes!

fmitchell-r7:
We probably want to document the new access log options in the [configuration section of the README](https://github.com/rapid7/propsd/blob/fd22e2ac6c2d1dc8f24aba6be12524de4e9fd1fe/README.md#configuration).

lib/logger.js
```diff
@@ -1,27 +1,46 @@
...
-    level: config.get('log:level'),
+    level: logLevel,
     transports: [
       new Winston.transports.Console({
         colorize: true,
         timestamp: true
       })
-    ]
+    ],
+    exitOnError: false
```
jmanero-r7: This only applies to caught unhanded-exceptions, which the logger should not be listening for. In any case, the process _should_ crash in the event of an unhanded-exception allowing Upstart to restart it.

lib/logger.js
```diff
@@ -1,27 +1,46 @@
...
   });
 
-  const filename = config.get('log:filename');
-
   if (filename) {
-    logger.add(Winston.transports.File, {
-      filename
-    });
+    logger.add(Winston.transports.File, {filename});
```
jmanero-r7: File logging from the service process is bad. All output should be sent out STDOUT or another streaming channel to something that will handle writing to disk and rotation correctly.

dgreene-r7: Not sure I follow. Are you saying that we should only use Winston to log to STDOUT and deal with writing to files outside of this project?

jmanero-r7: Yes. Because logging through upstart doesn't require any extra log-rotation configurations, doesn't require any integration with the logging module for file-truncation, and really, Just Works :tm:.

fmitchell-r7: I'm kind of in favor of the "log everything to STDOUT" argument, if only because it keeps the propsd code simpler. If we continue to support file logging within propsd, we'll want to fix #101 and expose log file rotation settings in the configuration.

If we log everything to STDOUT, we should allow users to turn access logging on/off in the configuration.

dgreene-r7: In that case we just don't provide a `log:access:filename` (or `log:filename`) config option. The log service only logs to a file if a filename option is provided.

jmanero-r7: I think we're overcomplicating this. We should:

1. Remove all file-logging from the config/logging modules. If other users want to log to file, they should `> to_a_file`, or `| to_a_log_handler`. Systemd, upstart, runit, and daemontools all do this by default, or with minimal configuration.

2. Log API requests in the default log-stream. There will not be very many of these. Only local resources have access to the propsd API.

dgreene-r7: I'm good with that.

fmitchell-r7:
Thank you for adding access logging!

I agree with the move to STDOUT as the canonical log source. But I'm concerned about the loss of functionality with the removal of file logging. I want `log:filename` to be a deprecated feature. If it's provided, add a transport for it, and issue a warning letting users know it's deprecated and will be removed in a future version.

Does that sound reasonable?

dgreene-r7:
I think that's very reasonable and something we should be considering for every feature change.

fmitchell-r7:
:+1: Thanks for adding the deprecation warning and including the test to make sure it's there! :shipit:

PR #100, opened 2016-03-15
Removed unnecessary .gitkeep files

dgreene-r7:


fmitchell-r7:
:+1: Thanks for cleaning these up! :shipit:

Issue #103, opened 2016-03-15
Consul tags should be explicit about cluster name


fmitchell-r7:
Right now, propsd [assumes services in Consul may or may not be tagged](https://github.com/rapid7/propsd/blob/f9a3baa811fb24be75a5041597b0cbe116de2997/lib/source/consul.js#L144-L169). If the service has tags, the tags are used as the cluster name. If the service doesn't have tags, the service name is used as the cluster name.

This messes things up if I'm using "elasticsearch" as a service name and "version:2.2" as a tag. Because my output property is `conqueso.version:2.2.ips` when it should be `conqueso.elasticsearch.ips`.

Instead, we should assume that clustered services in Consul are tagged as "cluster:name". If a service has a tag that starts with "cluster:", everything after "cluster:" should be used as the cluster name. If a service doesn't have a tag that starts with "cluster:", the service name should be used as the cluster name.

If multiple "cluster:" tags are defined for a service, we should end up with multiple output properties. So a service named "elasticsearch" with tags "cluster:logs" and "cluster:development" would output two properties: `conqueso.logs.ips` and `conqueso.development.ips`.

dgreene-r7:
The scope of this issue covers #69 and #70.

dgreene-r7:
Per conversations with @fmitchell-r7, @athompson-r7, and @eciramella-r7 tag-based watcher creation shouldn't be supported at this time. Services should map directly to conqueso roles and tags are ignored for now. Future versions after the conqueso transition may support arbitrary k:v tags for watcher creation and segmentation.

Issue #70, opened 2016-03-08
Refactor Consul plugin so service list parsing is in its own class


fmitchell-r7:
Looking at the S3 plugin, it [has a parser class](https://github.com/rapid7/propsd/blob/2154e00472f33e449a128232b6075edcce97dc54/lib/source/s3.js#L9-L24) that's responsible for formatting the byte structure returned from the AWS SDK into properties. This splits the separate of concern for parsing S3 data vs. retrieving S3 data into separate classes.

The Consul plugin should be refactored so it has similar behavior. Right now the parsing logic for service list watchers is [embedded in the fetch callback](https://github.com/rapid7/propsd/blob/2154e00472f33e449a128232b6075edcce97dc54/lib/source/consul.js#L201-L215). This should get moved into its own parser class.

PR #104, opened 2016-03-16
Fix: refactored service list watcher and service health watcher parsing into their own classes

### LINKED ISSUE #69: Refactor Consul plugin so health watcher parsing is in its own class
Looking at the S3 plugin, it [has a parser class](https://github.com/rapid7/propsd/blob/2154e00472f33e449a128232b6075edcce97dc54/lib/source/s3.js#L9-L24) that's responsible for formatting the byte structure returned from the AWS SDK into properties. This splits the separate of concern for parsing S3 data vs. retrieving S3 data into separate classes.

The Consul plugin should be refactored so it has similar behavior. Right now the parsing logic for health watchers is [embedded in the fetch callbacks](https://github.com/rapid7/propsd/blob/2154e00472f33e449a128232b6075edcce97dc54/lib/source/consul.js#L201-L215). This should get moved into its own parser class.

dgreene-r7:
This PR is the first of a few to cover #103. It focuses on refactoring watcher data parsing into their own classes, prefixed with `ConquesoCompat` to indicate their intermediate status. When consul tagging is complete we can refactor that class (or create new ones that conform to the interface) to migrate to our new methodology.

This PR resolves #69 and #70.

### DIFF
--- a/lib/source/consul.js
+++ b/lib/source/consul.js
@@ -31,6 +31,47 @@ class ConquesoCompatHealthParser {
     this.properties = properties;
   }
 }
+
+class ConquesoCompatServiceParser {
+  constructor() {
+    this.properties = {};
+  }
+
+  update(data, callback) {
+    const services = Object.keys(data);
+    const watcherInfo = [];
+
+    services.forEach((service) => {
+      const tags = data[service];
+
+      tags.forEach((tag) => {
+        watcherInfo.push({
+          name: `${service}-${tag}`,
+          cluster: tag,
+          options: {
+            passing: true,
+            service,
+            tag
+          }
+        });
+      });
+
+      if (tags.length <= 0) {
+        watcherInfo.push({
+          name: service,
+          cluster: service,
+          options: {
+            passing: true,
+            service
+          }
+        });
+      }
+    });
+
+    this.properties.watcherInfo = watcherInfo.filter(callback);
+  }
+}
+
 class Consul extends EventEmitter {
   /**
    * Creates a new instance of a Consul souce plugin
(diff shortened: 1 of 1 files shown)


lib/source/consul.js
```diff
@@ -7,6 +7,71 @@ const consul = require('consul');
...
+  constructor() {
+    this.properties = {};
+  }
+
+  update(data, callback) {
+    const services = Object.keys(data);
+    const watcherInfo = [];
+
+    services.forEach((service) => {
+      const tags = data[service];
```
jmanero-r7: This doesn't solve for a tagging scheme of key-value pairs like `["version:1.2.3", "cluster:foo"]`. We need to check if there's a `key:value` format, then only pick `cluster:...`, or if there are no tags matching that format, assume that all tags are cluster names.

dgreene-r7: That's coming in a later PR :smile:

lib/source/consul.js
```diff
@@ -7,6 +7,71 @@ const consul = require('consul');
 const DEFAULT_AGENT_ADDRESS = '127.0.0.1';
 const DEFAULT_AGENT_PORT = 8500;
 
+class ConquesoCompatHealthParser {
+  constructor() {
+    this.properties = {};
+  }
+
+  update(data, cluster) {
```
fmitchell-r7: Would you be willing to put in API docs around the `update` method?

lib/source/consul.js
```diff
@@ -7,6 +7,71 @@ const consul = require('consul');
...
+    this.properties = properties;
+  }
+}
+
+class ConquesoCompatServiceParser {
+  constructor() {
+    this.properties = {};
+  }
+
+  update(data, callback) {
```
fmitchell-r7: Would you be willing to put in API docs around the `update` method?

fmitchell-r7:
:+1: Thanks for doing this refactor! This should resolve #69 and #70 as well. :shipit:

PR #105, opened 2016-03-17
Consul source only uses service name

### LINKED ISSUE #103: Consul tags should be explicit about cluster name

dgreene-r7:
Per #103, we are removing tag parsing from the Consul source completely. Later versions will likely support a `key:value` format in tags. This PR resolves #103.

### DIFF
--- a/test/consul.js
+++ b/test/consul.js
@@ -460,4 +460,8 @@ describe('Storage engine', () => {
       }
     });
   });
+
+  it('resolves multiple tags as separate services');
+
+  it('avoids resolving similar tags to the same service');
 });
--- a/test/consul.js
+++ b/test/consul.js
@@ -346,57 +346,6 @@ describe('Consul', () => {
     });
   });
 
-  it('resolves multiple tags as separate services', () => {
-    const consul = generateConsulStub();
-
-    consul.initialize();
-    consul.mock.emitChange('catalog-service', {consul: ['production', 'development']});
-    consul.mock.emitChange('consul-production', [{
-      Service: {Address: '127.0.0.1'}
-    }]);
-    consul.mock.emitChange('consul-development', [{
-      Service: {Address: '10.0.0.0'}
-    }]);
-
-    return Promise.resolve(consul.properties.consul).should.eventually.eql({
-      'consul-production': {
-        cluster: 'production',
-        addresses: ['127.0.0.1']
-      },
-      'consul-development': {
-        cluster: 'development',
-        addresses: ['10.0.0.0']
-      }
-    });
-  });
-
-  it('avoids resolving similar tags to the same service', () => {
-    const consul = generateConsulStub();
-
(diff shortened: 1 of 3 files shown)


dgreene-r7:
Thanks, coveralls, for failing this PR because it decreased coverage by ~0.0009%.

test/consul.js
```diff
@@ -346,57 +346,6 @@ describe('Consul', () => {
     });
   });
 
-  it('resolves multiple tags as separate services', () => {
```
fmitchell-r7: Is this test case (agents with different IP addresses and tags registering the same service should resolve all the addresses to the service) covered elsewhere? If not, we might want to fix it so it passes instead of removing it.

test/consul.js
```diff
@@ -346,57 +346,6 @@ describe('Consul', () => {
...
-        addresses: ['127.0.0.1']
-      },
-      'consul-development': {
-        cluster: 'development',
-        addresses: ['10.0.0.0']
-      }
-    });
-  });
-
-  it('avoids resolving similar tags to the same service', () => {
```
fmitchell-r7: Is this test case (agents with different IP addresses and the same tag registering to different services should resolve the addresses to each individual service) covered elsewhere? If not, we might want to fix it so it passes instead of removing it.

dgreene-r7: I think I'll set this as a "pending" test. We can fill in the implementation when we include tag parsing.

test/consul.js
dgreene-r7: I think I'll set this as a "pending" test. We can fill in the implementation when we include tag parsing.

fmitchell-r7:
:+1: Thanks for fixing this! Because this changes the Consul behavior relative to the 1.0.0 release, we'll want to make sure and include release notes indicating what users need to be aware of when upgrading. :shipit:

PR #106, opened 2016-03-21
Improve test coverage

dgreene-r7:
Refactored a few places to make testing simpler as well.

### DIFF
--- a/test/core-api-v1.js
+++ b/test/core-api-v1.js
@@ -147,4 +147,8 @@ describe('Core API v1', () => {
         done();
       });
   });
+
+  it('returns a 500 if any source plugins fail');
+
+  it('returns a 429 if any source plugins have a warning');
 });
--- a/lib/source/metadata.js
+++ b/lib/source/metadata.js
@@ -204,11 +204,13 @@ class Metadata extends EventEmitter {
     /**
      * Initialize the metadata-service client
      */
+    const host = options.host || Config.get('metadata:host');
+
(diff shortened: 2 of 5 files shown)


test/core-api-v1.js
```diff
@@ -147,4 +147,8 @@ describe('Core API v1', () => {
         done();
       });
   });
+
+  it('returns a 500 if any source plugins fail');
+
+  it('returns a 429 if any source plugins have a warning');
```
fmitchell-r7: Is there a commit that fills in these tests?

dgreene-r7: Not yet. The spec around how we handle source statuses is incomplete.

Right now the `ok` property in the object returned from `Source#status()` is a bool. Eventually we'll want something to determine whether it's in a `warn` state and once we figure out how to handle that, we can implement it in the sources and fill in these tests.

I *think* this only affects a consul health check for propsd and shouldn't be critical but I'm not sure. The answer is probably deep within the guts of the conqueso client.

fmitchell-r7: :+1:

fmitchell-r7:
Thanks for improving out test coverage numbers! :shipit:

test/core-api-v1.js
dgreene-r7: I filed the need to complete the spec to define how we'll write the code that these tests will test (whew, that was a mouthful) as issue #107.

PR #109, opened 2016-03-21
Fix: Travis test timeout latency

dgreene-r7:
Travis introduces enough lag in test execution time that at least one of our tests is failing because it times out and a few others are getting close to the timeout duration. This PR bumps that to 5 seconds to give us a bit of a buffer.


fmitchell-r7:
:+1: Thanks for fixing this!

PR #112, opened 2016-03-24
Cookbook

jmanero-r7:
This PR adds a cookbook for installing and configuring propsd from a published DEB artifact.

It requires a wrapping cookbook to define implementation-specific service configurations, and for the service resource to have an action set: `resources('service[propsd]').action(:start, :enable)`


Gemfile.lock
```diff
@@ -0,0 +1,239 @@
+GEM
```
erran-r7: @jmanero-r7 are we sure we want to commit Gemfile.lock? Standard practice is that you commit it for application code, but not for library code. [This blog post](http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/) summarises the reasons why.

erran-r7: @jmanero-r7 disregard that message. :sweat_smile:

erran-r7:
:fire: :fire: :fire: Now we're cooking with gas. :fire: :fire: :fire: Looks good to me.

dgreene-r7:
LGTM :shipit:

Issue #102, opened 2016-03-15
Project: Getting Started Guide


fmitchell-r7:
We need to provide some documentation around how to get started with propsd. The documentation should be aimed at people who want to install propsd on a server and use it to deliver configuration. A basic working knowledge of Linux and Amazon's S3 service can be assumed. The getting started documentation can cover the following:

* Installing propsd (see #115)
  * On Debian systems
  * From source
* Writing a configuration file for propsd (see #116)
  * Default options for the HTTP server and how to change them
  * Default options for logging and how to change them
  * Default options for S3 polling and how to change them
* Starting the propsd service (see #117)
* Checking that the propsd service is running (see #117)
* Setting up a S3 bucket for propsd
  * What permissions the bucket and instance need
  * The JSON file format for files in S3
  * Configuring propsd to read from the bucket
  * How properties in the JSON file map to Conqueso output

Advanced topics (these may need to go into other guides) can cover the following:

* Layering properties
  * Loading multiple files as sources from S3
  * How properties are merged from multiple sources
  * Interpolated strings in properties
  * An example layering scheme
* Integrating Consul with propsd
  * Setting up a Consul health check for propsd
  * Using Consul has a source for properties
  * How Consul properties map to Conqueso output

athompson-r7:
Can we limit the scope of `non-Debian` systems?  i.e. Redhat-based, from source, etc.

fmitchell-r7:
@athompson-r7 Good call. I changed it to be "Debian" and "from source" since those are the too formats we have released artifacts for.

dgreene-r7:
@fmitchell-r7 I think we can close this. Wanna do the honors?

PR #118, opened 2016-04-12
Feature: Installation guide

### LINKED ISSUE #115: Write installation guide for propsd
We need to write an installation guide for propsd. This should cover installing propsd from a Debian package and installing propsd from source. The resulting Markdown file should live in "docs/getting-started/installation.md" under the root of the project. There should be a "README.md" file in the "docs/getting-started" folder that provides a table of contents with links to the installation sections.

fmitchell-r7:
This fixes #115. There's now a getting started guide for propsd in the "docs/getting-started" folder. The README has been updated to include a link to the getting started guide.

Currently the guide only covers installation. There are addition issues filed to cover the remaining bits like configuration and usage (see #102 for the complete scope of this project).


docs/getting-started/installation.md
```diff
@@ -0,0 +1,50 @@
+# How to install propsd #
+
+[Releases of propsd][releases] include both source tarballs and Debian
+packages. Debian and Ubuntu based Linux distributions can use the pre-built
+packages. Other operating systems should install propsd from source.
+
+## Installing from the Debian package ##
+
+Propsd depends on Node.js, so follow the [instructions for installing Node.js
+on Debian based systems][node-debian].
```
athompson-r7: Which version?

docs/getting-started/installation.md
```diff
@@ -0,0 +1,51 @@
+# How to install propsd #
+
+[Releases of propsd][releases] include both source tarballs and Debian
+packages. Debian and Ubuntu based Linux distributions can use the pre-built
```
eciramella-r7: Is it worth linking to latest (https://github.com/rapid7/propsd/releases/latest)?

Curious if users would just click through things during the reading of the installation guide.

docs/getting-started/installation.md
```diff
@@ -0,0 +1,51 @@
...
+`/tmp/propsd.tar.gz`.
+
+Create a new folder for propsd. These instructions assume you're using
+`/opt` as that folder.
+
+~~~bash
+mkdir /opt
+~~~
+
+Use `npm` to install propsd.
```
dgreene-r7: We should probably publish to npm...

fmitchell-r7: Good idea! Let's open an issue for that so we a) don't forget, and b) can have a larger conversation around it.

docs/getting-started/installation.md
fmitchell-r7: I update the "releases" related links so they now go to the latest release instead of the "look, here's every possible release" page. That should make it easier to get started when you're reading and clicking through the guide.

docs/getting-started/installation.md
dgreene-r7: Filed as #120

PR #124, opened 2016-04-14
Bug: Deleted config file doesn't trigger S3 source shutdown

### LINKED ISSUE #123: Deleted config file doesn't trigger S3 source shutdown
When using transient config files (i.e. an instance-specific config for a canary node), deleting the config file doesn't trigger the source shutdown.

dgreene-r7:
This PR updates `_fetch()` to shut the source down and clear its properties if S3.getObject returns a NoSuchKey error. Resolves #123.

### DIFF
--- a/lib/source/s3.js
+++ b/lib/source/s3.js
@@ -228,7 +228,6 @@ class S3 extends EventEmitter {
           return callback(null, false);
         }
         if (err.code === 'NoSuchKey') {
-          this.shutdown();
           return callback(null, {properties: {}});
         }
 
--- a/lib/source/s3.js
+++ b/lib/source/s3.js
@@ -228,7 +228,7 @@ class S3 extends EventEmitter {
           return callback(null, false);
         }
         if (err.code === 'NoSuchKey') {
-          return callback(null, {properties: {}});
+          return callback(null, new Buffer(JSON.stringify({properties: {}})));
         }
 
         return callback(err);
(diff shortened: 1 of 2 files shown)


fmitchell-r7:
Does the `PluginManager` need to do something with the "shutdown" event? Right now it looks like [it's only logging a message](https://github.com/rapid7/propsd/blob/bug-removed-config-file-should-trigger-s3-source-shutdown/lib/plugin-manager.js#L257-L259). Should we clear the storage and unregister the plugin as well?

jmanero-r7:
I don't think that this is the correct behavior. The service should continue to check for non-existent objects because they may be added in the future.

dgreene-r7:
I agree but isn't that handled by the index checking its manifest against the list of sources?

fmitchell-r7:
What @jmanero-r7 said. I think we can just clear the properties. As far as I can see, there's no need to shut down the source. It will either get destroyed on its own when the index changes to remove it, or it will have new properties when a new object is created in S3.

jmanero-r7:
> dgreene-r7 commented 5 minutes ago
> I agree but isn't that handled by the index checking its manifest against the list of sources?

This should only be triggered by an update to the index object is updated.

dgreene-r7:
> This should only be triggered by an update to the index object is updated.

Yup, thanks for reminding me. I forgot that we're no-oping `NotModified` errors from the S3 client.

So the correct behavior that we're looking for is that the plugin sticks around but its `properties` are cleared?

fmitchell-r7:
Thanks for fixing this!
:+1: :shipit:

PR #128, opened 2016-05-11
Expose propsd version in cookbook node attribute

dgreene-r7:
We need to get the version of propsd we're deploying further upstream where `cookbook_version.version` will return something other than the propsd version. This PR adds `version` to the `propsd` attributes so we can use it later.


jmanero-r7:
👍

PR #130, opened 2016-05-13
Fixed issue where cookbook_version was used incorrectly.

dgreene-r7:


simonirwin-r7:
LGTM, @dgreene-r7 I just hit this same issue in one of the Luxor apps

PR #133, opened 2016-05-17
Feature: Add version property to health endpoint

dgreene-r7:
This PR adds the `version` property to the `/v1/health` endpoint which outputs the application version.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -77,7 +77,7 @@ task :package_dirs do
 end
 
 task :propsd_source => [:install] do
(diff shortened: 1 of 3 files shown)


rvanburen-r7:
👍

fmitchell-r7:
Thanks for adding this!

Would you mind writing a test to validate that the version is included in the output from the `/v1/health` endpoint?

fmitchell-r7:
👍  Thanks for adding the test. :shipit:

PR #137, opened 2016-05-20
Adding some more color to the README page.

eciramella-r7:
From the README, it's tough to understand what this thing does or what the benefits are.  Trying to simplify it without redoing everything that's on/in the gsg page.


fmitchell-r7:
👍 Having a feature list in the README is a good idea. Thanks for adding this. :shipit:

PR #139, opened 2016-05-20
Feature: Get specific property

### LINKED ISSUE #136: Conqueso API for reading specific properties isn't supported
Users that are upgrading from Conqueso expect to have an compatible API. We should support reading specific properties e.g. the [GET /api/roles/:role/properties/:property](https://github.com/rapid7/conqueso#get-apirolesrolepropertiesproperty) API. This can be an enhancement to the `/v1/conqueso` API, so we don't  need a new `/v2` API for this.

dgreene-r7:
This PR resolves #136. It only retrieves first-level properties.

For example, `/v1/conqueso/foo.bar` will return the value at `storage.properties['foo.bar']` whereas `/v1/conqueso/foo/bar` will return the value at `storage.properties['foo']['bar']`.

### DIFF
--- a/lib/control/v1/conqueso.js
+++ b/lib/control/v1/conqueso.js
@@ -110,6 +110,21 @@ function makeConquesoProperties(properties) {
  * @param {Storage} storage
  */
 function Conqueso(app, storage) {
+  // Conqueso compatible APIs are defined before the generic catch all route.
+  app.get('/v1/conqueso/api/roles/:role/properties/:property', (req, res) => {
+    const property = req.params.property;
+
+    res.set('Content-Type', 'text/plain');
+    res.status(HTTP_OK);
+
+    if (property && (property in storage.properties)) {
+      res.end(String(storage.properties[property]));
+    } else {
+      res.end();
+    }
+  });
+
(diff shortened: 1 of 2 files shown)


lib/control/v1/conqueso.js
```diff
@@ -120,9 +120,19 @@ function Conqueso(app, storage) {
   }
 
   route.get((req, res) => {
+    const prop = req.params['0'].replace(new RegExp('/', 'g'), '');
```
jmanero-r7: I think there's a better way to handle this. Instead of writing your own regexp, add handlers for both the URI generated by the init script and the single-property endpoint, then ignore any unnecessaery dynamic URI parameters:

```javascript
app.get('/v1/conqueso', all_properties)
app.get('/v1/conqueso/api/:role/properties', all_properties)
app.get('/v1/conqueso/api/:role/properties/:property', single_property)
```

dgreene-r7: There definitely is and your way is preferable. I had to poop this out in a short period of time as it was blocking @storance-r7.

jmanero-r7:
👍

PR #141, opened 2016-05-25
Fix: Cache properites when index rebuilds

### LINKED ISSUE #140: Storage service isn't caching properties when the index changes
The [`Storage` object clears out properties](https://github.com/rapid7/propsd/blob/v1.2.3/lib/storage.js#L84) when [the index changes](https://github.com/rapid7/propsd/blob/v1.2.3/lib/plugin-manager.js#L193). This means there's a race condition. If you happen to poll for properties when `Storage` has been cleared but before the new index has been parsed, you'll get back empty properties.

We don't need to clear the properties when unregistering sources. The [`Storage#update` call](https://github.com/rapid7/propsd/blob/v1.2.3/lib/storage.js#L63-L73) takes care of rebuilding the cache.

fmitchell-r7:
This fixes #140. Properties are no longer cleared when sources are unregistered. Instead, we'll persist properties in memory. In memory properties will be updated when the underlying sources finish parsing and update.

### DIFF
--- a/lib/storage.js
+++ b/lib/storage.js
@@ -81,7 +81,6 @@ class Storage {
     while (i--) {
       this.unregister(this.sources[i]);
     }
-    this.properties = Object.create(null);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


davepgreene:
:+1: 

Thanks for the quick fix!

PR #142, opened 2016-05-27
Feature: improve logging & bugfixes

dgreene-r7:
This PR fixes a bug where mutations were occurring when the `storage.properties` object was being cloned by reference via the structured clone algorithm whenever a GET request went to the `/v1/conqueso/*` endpoint.

Second, it adds a check if the storage properties are rebuilding and if so, the `/v1/conqueso/*` endpoints return a `503` status code so that Archaius will use the previously stored value.

Finally, it adds logging around when updates that carry a change are assigned to `storage.properties` to hopefully track down bugs in software that consumes propsd.


lib/storage.js
```diff
@@ -69,6 +72,13 @@ class Storage {
...
       merge(source.properties, properties);
     });
+
+    const diff = deepDiff(this.properties, properties);
+
+    if (diff) {
+      Log.info(`Calculated property diff`, {diff});
+    }
+
     this.properties = properties;
```
jeffreymyers-r7: Should this.properties only be updated if there's a diff?

lib/plugin-manager.js
```diff
@@ -251,20 +251,20 @@ class PluginManager extends EventEmitter {
    */
   _registerSourceEvents(source) {
     source.on('startup', () => {
-      Log.verbose(`${source.name} started up.`);
+      Log.debug(`${source.name} started up.`);
```
jeffreymyers-r7: Consider logging startup and shutdown at an info level

lib/plugin-manager.js
```diff
@@ -251,20 +251,20 @@ class PluginManager extends EventEmitter {
...
     });
 
     source.on('shutdown', () => {
-      Log.verbose(`${source.name} shut down.`);
+      Log.debug(`${source.name} shut down.`);
     });
 
     source.on('update', () => {
-      Log.verbose(`${source.name}'s data was updated from its underlying source data.`);
+      Log.debug(`${source.name}'s data was updated from its underlying source data.`);
```
jeffreymyers-r7: Consider logging this at an info level

fmitchell-r7:
Thanks for tracking down this bug!

Is there any way we can write a test for this? It'd be nice to avoid updating `clone` or replacing it with another library in the future and reintroducing this issue.

lib/storage.js
dgreene-r7: You're right but I'd like to understand the diffs before we confirm that properties should only be updated if there is a diff as calculated by this lib. I'm adding more diff logging so we can get some insight into the current behavior.

jeffreymyers-r7: Alright. I'm all for insight. :dollar:

dgreene-r7:
> Is there any way we can write a test for this? It'd be nice to avoid updating clone or replacing it with another library in the future and reintroducing this issue.

I'm not entirely sure. Our best bet is probably to fix the library version or port their code into propsd. I'm amenable to either. Preferences?

lib/plugin-manager.js
```diff
@@ -250,46 +250,55 @@ class PluginManager extends EventEmitter {
...
     this._ok = false;
 
+    const logMetadata = (source) ? {source: source.name, type: source.type} : {};
+
     // Only emit an error event if there are listeners.
     if (this.listeners('error').length > 0) {
-      this.emit('error', err);
+      this.emit('error', err, logMetadata);
+    } else {
+      Log.error(err, logMetadata);
```
jmanero-r7: This shouldn't be conditional

lib/source/consul.js
```diff
@@ -343,19 +342,19 @@ class Consul extends EventEmitter {
...
 
-    Log.error(error, {
-      source: this.name,
-      type: this.type
-    });
-
     if (this.listeners('error').length > 0) {
-      this.emit('error', error);
+      this.emit('error', err);
+    } else {
```
jmanero-r7: This shouldn't be conditional

lib/control/v1/conqueso.js
```diff
@@ -137,7 +116,7 @@ function Conqueso(app, storage) {
 
   route.get((req, res) => {
     res.set('Content-Type', 'text/plain');
-    res.status(HTTP_OK);
+    res.status(statusCode);
     res.end(makeJavaProperties(makeConquesoProperties(storage.properties)));
```
jeffreymyers-r7: Can we avoid sending the body if statusCode != HTTP_OK?

fmitchell-r7:
I'm trying to wrap my head around what's in the PR.

* clone function was replaced with [a library](https://github.com/pvorb/node-clone)
* INFO level logging around start, stop, update moved from sources to plugin manager
* verbose logging was added for deltas when plugins update
* 503 is returned (instead of 200) while properties are being computed and updated

@dgreene-r7, does that sum it up?

dgreene-r7:
> I'm trying to wrap my head around what's in the PR.
> 
> clone function was replaced with a library
> INFO level logging around start, stop, update moved from sources to plugin manager
> verbose logging was added for deltas when plugins update
> 503 is returned (instead of 200) while properties are being computed and updated
> @dgreene-r7, does that sum it up?

Yup, that's everything. Normally I'd have split it out into multiple PRs but that urgency tho.

rvanburen-r7:
I think this makes sense. though you probably want more eyes on this.

jmanero-r7:
:shipit:

PR #143, opened 2016-05-27
Split the source/metadata module into a few submodules for readability

jmanero-r7:


lib/source/metadata.js
```diff
@@ -8,166 +8,11 @@ const Path = require('path');
...
-        'local-ipv4s', 'interface-id'
-      ].forEach((name) => interfaceProperties[name] = data[Path.join(interfacePath, name)]);
-    }
-
-    this.properties = {};
-    this.properties[this.namespace] = properties;
-  }
-}
+const Parser = require('./metadata/parser');
+const Util = require('metadata/util');
```
dgreene-r7: Shouldn't this be `require('./metadata/util')`?

jmanero-r7: Yes

dgreene-r7:
:shipit:

PR #145, opened 2016-05-27
Make `Source#initialize` return a Promise

jmanero-r7:
This will simplify the definition of dependency chains in the manager module.

* Remove the `Metadata#clear` method. This operation souldn't be possible.
* Move more a lot of boilerplate from the Metadata module into a common base for Source classes

### DIFF
--- a/lib/source/common.js
+++ b/lib/source/common.js
@@ -105,9 +105,6 @@ const Source = (Parser) => class extends EventEmitter {
       return this;
     }
 
-    this._ok = false;
-    this.signature = null;
-
     Log.info(`Stopping down ${this.type} source ${this.name} polling`, {
       source: this.name,
       type: this.type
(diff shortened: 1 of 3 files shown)


fmitchell-r7:
👍 Thanks for refactoring this! :shipit:

lib/source/common.js
```diff
@@ -0,0 +1,184 @@
+/* global Log, Config */
+'use strict';
+
+const EventEmitter = require('events').EventEmitter;
+const DEFAULT_INTERVAL = 60000;
+
+const deepDiff = require('deep-diff').diff;
+
+const Source = (Parser) => class extends EventEmitter {
```
dgreene-r7: It took me a few seconds to realize what this was doing but damn that's awesome.

jmanero-r7: Shamelessly boosted that idiom from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#Mix-ins

dgreene-r7:
Looks great!

PR #146, opened 2016-05-27
Fix: Better logic around source reloading

dgreene-r7:
This PR improves the way we handle source reloading. Instead of removing all sources and rebuilding the index, this adds sources if they aren't in `storage.sources` and are in the new index and removes them if they're in `storage.sources` but aren't in the new index.

This is a naive solution and can definitely be improved.

### DIFF
--- a/lib/plugin-manager.js
+++ b/lib/plugin-manager.js
@@ -185,44 +185,73 @@ class PluginManager extends EventEmitter {
   }
 
   /**
+   * Factory function to create instances of source plugins
+   * @param {object} source
+   * @returns {Metadata|S3|Consul}
+   * @private
+   */
+  _sourceFactory(source) {
+    let instance;
+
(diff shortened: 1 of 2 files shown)


lib/plugin-manager.js
```diff
@@ -185,44 +185,73 @@ class PluginManager extends EventEmitter {
...
+          host: Config.get('consul:host'),
+          port: Config.get('consul:port'),
+          secure: Config.get('consul:secure')
+        });
+        break;
+      default:
+        this._error(new Error(`Source type ${source.type} not implemented`));
+    }
+
+    if (instance) {
```
jmanero-r7: This doesn't really do anything. If you don't return a falsey value from `instance`, the call will return undefined anyway

jmanero-r7:
👍 Looks good

PR #148, opened 2016-05-27
Enhancement: Log granularity

dgreene-r7:
This PR adds some more information to logging. Ideally after #133 is merged we can refactor it to allow application information to be exposed globally so we can log more things.


fmitchell-r7:
Thanks for fixing this!

It looks like we're logging local properties (from the config) for every source start/stop since that's the default metadata. Is that by intention? It feels like it might better to just log those once on service start, since they aren't likely to change.

dgreene-r7:
Perhaps only on service start and index update?

fmitchell-r7:
Do we want to change logging in the [`Common` source class](https://github.com/rapid7/propsd/blob/master/lib/source/common.js#L37-L40) as well?

jmanero-r7:
Rebase Plz

lib/plugin-manager.js
```diff
@@ -251,55 +265,77 @@ class PluginManager extends EventEmitter {
...
+        Object.assign({
+          sourceName: source.name,
+          sourceType: source.type
+        },
+        logMetadata));
     });
 
-    source.on('startup', () => {
-      Log.info(`${source.name} started up successfully.`);
+    source.on('startup', (logMetadata) => {
```
jmanero-r7: The `init` and `startup` events are fired from the same closures in all source modules. We should pick one and remove the other.

dgreene-r7: Agreed, same with `stop` and `shutdown` unless we want to distinguish behaviors.

lib/logger.js
```diff
@@ -2,12 +2,43 @@
...
+ * @param {string} filename
+ * @returns {Winston.Logger}
+ * @constructor
+ */
 function Logger(level, filename) {
-  const logLevel = level || 'info';
+  const logLevel = level.toUpperCase() || 'INFO';
+
+  /* eslint-disable rapid7/static-magic-numbers */
+  const javaLogLevels = {
```
erran-r7: Java has several log utilities and "SILLY"/"VERBOSE" aren't levels I know in any. Seems like these come from Winston. Should we rename this variable to `customLogLevels` since that's what we're defining here?

dgreene-r7: Yeah, [Logback](http://logback.qos.ch/apidocs/ch/qos/logback/classic/Level.html) uses (not listed in order of severity) `DEBUG`, `ERROR`, `INFO`, `WARN`, AND `TRACE`. The problem is that `SILLY` and `VERBOSE` are used for two different things but map more closely to `TRACE`.

Agreed that the var should be renamed. Maybe `javaLikeLogLevels`? 😝

erran-r7: ❤️

test/bin/s3-server.js
```diff
@@ -42,7 +42,7 @@ const port = args.p;
 const path = Path.resolve(__dirname, `../../${args.d}`);
 
 nconf.set('log:level', 'debug');
-const Log = require('../../lib/logger').attach(nconf);
+const Log = require('../../lib/logger').attach('debug');
```
fmitchell-r7: Why the removal of configuration options to the logger?

fmitchell-r7:

PR #149, opened 2016-05-27
Fix: Keep index sources ordered

fmitchell-r7:
This reverts the changes introduced by #146. We accidentally introduced a bug where the sources defined in the index could get out of order. A test has been added to ensure the sources in the index stay in the order they're defined.


dgreene-r7:
:shipit: but :(

PR #151, opened 2016-05-30
Refactor Source to be a better defined state machine.

jmanero-r7:
* Introduces a `#state` property and named states which consolidate previous `running` and `ok`
  registers.
* Introduces an additional return value for `#_fetch` to indicate that a source is `non-existent`,
  not just `not-updated`. If the source was RUNNING, this signal returns the source to INITIALIZING
  and notifies watchers accordingly.

Other Changes:

* Separate polling controls from initialize/shutdown
* Catch errors thrown from parsers
* Allow Sources' default Parsers to be overridden by `options.parser`

### DIFF
--- a/lib/source/common.js
+++ b/lib/source/common.js
@@ -91,7 +91,6 @@ const Source = (Parser) => class extends EventEmitter {
       Log.debug(`Polling source ${this.name} for updates`, {
         source: this.name, type: this.type});
 
-
       this._fetch((err, data) => {
         if (!this._timer) {
           // If `stop()` was called _during_ a fetch operation, `clearTimeout()`
--- a/lib/source/common.js
+++ b/lib/source/common.js
@@ -10,63 +10,97 @@ const Source = (Parser) => class extends EventEmitter {
   constructor(options) {
     super();
 
-    this._parser = new Parser(options);
+    this._parser = options.parser || new Parser(options);
 
     this.type = options.type;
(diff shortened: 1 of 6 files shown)


dgreene-r7:
👍  Looks good. I assume fully promisify-ing the sources comes later?

fmitchell-r7:
👍  Thanks for cleaning this up!

I think there's [a comment](https://github.com/rapid7/propsd/blob/source-cleanup/lib/source/common.js#L37-L38) that doesn't match the code, since the promise now resolves on "initialize" instead of "update". Other than that, :shipit:.

dgreene-r7:
What's the after-work for this PR? Is it just refactoring the `plugin-manager` or are there other effects elsewhere?

jmanero-r7:
#153 implements the logic to add/remove `update` listeners after all sources' initialize promises resolve, then some work to have the plugin-manager that.

dgreene-r7:
Awesome. I'd suggest @fmitchell-r7 take a quick look again but it LGTM.

fmitchell-r7:
👍  Looking at [code coverage](https://coveralls.io/builds/6418858), it looks like we're missing some of the test cases around state transition e.g initialized to waiting, warning to waiting, etc. But don't let that stop you from shipping this. 🚢

jmanero-r7:
Yeah, we need to write more targeted tests for the common module... TODO

Issue #147, opened 2016-05-27
Properties are missing while the index is being rebuilt


fmitchell-r7:
When the index is rebuilt, all of the [sources are reinitialized](https://github.com/rapid7/propsd/blob/v1.2.4/lib/plugin-manager.js#L187-L227). During the window that [S3 sources](https://github.com/rapid7/propsd/blob/v1.2.4/lib/source/s3.js) are initializing, their update events will trigger rebuilds of the properties in `Storage` before all of the new sources have properties to be merged. This results in missing properties for a short window.

fmitchell-r7:
The [metadata source](https://github.com/rapid7/propsd/blob/v1.2.4/lib/source/metadata.js) triggers an index rebuild, and it will refresh every hour as an EC2 instance rolls its access keys. The first step to fixing this is wrap the metadata and index initialization in a `Promise` so they can be chained. This will guarantee that metadata finishes initializing before it's used to resolve interpolated properties in the index. See #145 for the refactoring work around the metadata source.

dgreene-r7:
See #146 for refactoring work around source reloading.

onefrankguy:
I'm starting to wonder if this is completely solvable. Consider the case where we have an index that references a single source a. We want to change the index to reference source b. That requires the following operations: uploading source b (POST), updating the index (PUT), and deleting source a (DELETE). Using the notation "OPERATION(s3) => propsd" to denote operations on objects in S3 and the state of properties in propsd after it's processed those operations, there are six possible permutations for property resolution.

1. POST(b) => a, PUT(index) => b, DELETE(a) => b
2. POST(b) => a, DELETE(a) => null, PUT(index) => b
3. PUT(index) => null, POST(b) =>  b, DELETE(a) => b
4. PUT(index) => null, DELETE(a) => null, POST(b) => b
5. DELETE(a) => null, PUT(index) => null, POST(b) => b
6. DELETE(a) => null, POST(b) => null, PUT(index) => b

(1) keeps properties in the expected state. (2) and (3) incur the least amount of time in a state with missing properties. (4), (5), and (6) incur the most time with missing properties.

dgreene-r7:
See #151 and #153 for additional work.

PR #152, opened 2016-05-31
Metadata source should only poll once

dgreene-r7:
This is a quick fix to prevent hourly updates to the ec2 metadata service from triggering a full source rebuild. Once we've resolved that issue, we can reimplement the metadata source as a polling source.

Ref #147

### DIFF
--- a/lib/plugin-manager.js
+++ b/lib/plugin-manager.js
@@ -96,6 +96,7 @@ class PluginManager extends EventEmitter {
     });
     this.metadata.on('error', (error) => {
       this._error(error);
+      this.metadata.once('update', () => this._loadSourcesFromIndex());
     });
 
     this._running = false;
--- a/lib/plugin-manager.js
+++ b/lib/plugin-manager.js
@@ -97,9 +97,6 @@ class PluginManager extends EventEmitter {
     this.metadata.on('error', (error) => {
       this._error(error);
     });
-    this.metadata.on('update', () => {
-      this._loadSourcesFromIndex();
-    });
 
     this._running = false;
     this._ok = true;
@@ -111,8 +108,9 @@ class PluginManager extends EventEmitter {
   initialize() {
     Log.info('Initializing index and metadata');
     this._running = true;
-    this.index.initialize();
-    this.metadata.initialize();
+    this.metadata.initialize().then(() => {
+      this.index.initialize();
+    });
   }
 
   /**
--- a/lib/plugin-manager.js
+++ b/lib/plugin-manager.js
@@ -90,8 +90,10 @@ class PluginManager extends EventEmitter {
       this._loadSourcesFromIndex();
     });
 
-    this.metadata = new Metadata();
-    this.metadata.service.host = configuration.metadataHost || Config.get('metadata:host');
+    this.metadata = new Metadata({
+      host: configuration.metadataHost || Config.get('metadata:host'),
+      interval: configuration.interval || Config.get('metadata:interval')
+    });
     this.metadata.on('error', (error) => {
       this._error(error);
     });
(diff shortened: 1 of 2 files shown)


fmitchell-r7:
I'd like to solve this in a slightly more reusable way, so we don't wind up with custom `start` functions for every source that's one shot. What about having the `Common` class [skip the reschedule](https://github.com/rapid7/propsd/blob/master/lib/source/common.js#L69-L70) of polling if the interval's negative?

lib/plugin-manager.js
```diff
@@ -90,8 +90,10 @@ class PluginManager extends EventEmitter {
       this._loadSourcesFromIndex();
     });
 
-    this.metadata = new Metadata();
-    this.metadata.service.host = configuration.metadataHost || Config.get('metadata:host');
+    this.metadata = new Metadata({
+      host: configuration.metadataHost || Config.get('metadata:host'),
+      interval: configuration.interval || Config.get('metadata:interval')
```
fmitchell-r7: You probably want to get `config.metadataInterval` here so you're not accidentally resolving this as the polling interval for the index.

lib/plugin-manager.js
```diff
@@ -90,8 +90,10 @@ class PluginManager extends EventEmitter {
       this._loadSourcesFromIndex();
     });
 
-    this.metadata = new Metadata();
-    this.metadata.service.host = configuration.metadataHost || Config.get('metadata:host');
+    this.metadata = new Metadata({
+      host: configuration.metadataHost || Config.get('metadata:host'),
+      interval: configuration.metadataHost || Config.get('metadata:interval')
```
fmitchell-r7: Looks like a copy paste error? I think this should be `configuration.metadataInterval`.

dgreene-r7: Yuuuuup

jmanero-r7:
This is going to conflict with main-line updates to source modules. What if we just replace https://github.com/rapid7/propsd/blob/master/lib/plugin-manager.js#L98-L100 with 

```
  /**
   * Initialize the PluginManager's index and metadata source
   */
  initialize() {
    Log.info('Initializing index and metadata');
    this._running = true;
    this.index.initialize().then(() => {
      this.metadata.initialize();
    });
  }

```

dgreene-r7:
@jmanero-r7 The update to source modules isn't done yet, right? This PR is just a quickfix. I fully expect us to refactor the `plugin-manager` once source implementations are updated.

jmanero-r7:
It's ready to merge now

dgreene-r7:
Awesome! Let me take a look. We can do this in a nice order 😃

jmanero-r7:
I still don't think we should do this. https://github.com/rapid7/propsd/pull/152#issuecomment-223028048 doesn't require further changes to any of the source modules and doesn't break any of their tests.

dgreene-r7:
This doesn't break any tests either and it gives us an outlet to define a source that uses a lot of the `common` features but is a one-off execution.

jmanero-r7:
We should accomplish that by separating the polling code into a child-class and making the core state-machine generic. We'll need to do this to update the Consul source anyway.

jmanero-r7:
:shipit:

fmitchell-r7:
Can we add a test case for the initialization order? Something like this should work.

```javascript
it('initializes the metadata before the index', function (done) {
  const order = [];

  function testInitializeOrder(source) {
    order.push(source);
    if (order.length >= 2) {
      order.should.eql(['metadata', 'index']);
      done();
    }
  }

  manager.index.once('initialized', () => {
    testInitializeOrder('index');
  });

  manager.metadata.once('initialized', () => {
    testInitializeOrder('metadata');
  });

  manager.initialize();
});
```

fmitchell-r7:

PR #153, opened 2016-05-31
Implement a module to manage properties from all of the active sources

jmanero-r7:
* Replaces the storage module


dgreene-r7:
I like the thought process but in lieu of tests or documentation, can you provide a functional example of use? I just want to get a sense for how this slots in.

fmitchell-r7:
I like the distinction between static and dynamic sources. I like the idea of delaying the merging of properties so we don't have more than one merge running if multiple sources update in a short time window.

I'm not sure what the `View` class is doing or who's responsible for calling `View#activate`.

lib/properties.js
```diff
@@ -0,0 +1,197 @@
...
+  get properties() {
+    return this.source.properties;
+  }
+}
+
+/**
+ * A View aggregates the initialized events of multiple sources and safely
+ * forwards their update events to a parent Properties object.
+ */
+class View extends Layer {
```
fmitchell-r7: Does `View` need to extend `Layer`? It doesn't look like it's emitting events or leveraging the namespace concept that `Layer` provides.

jmanero-r7: It does not. I'm committing up a lot of this right now

test/properties.js
```diff
@@ -0,0 +1,237 @@
...
+    const view = properties.view();
+    const stub = new Stub({
+      interval: 500
+    }, {
+      foo: 'bar'
+    });
+
+    view.register(stub);
+
+    properties.once('build', (props) => {
```
jmanero-r7: TODO: This tests the build-on-activate, not on-source-update

test/properties.js
```diff
@@ -0,0 +1,237 @@
...
+  });
+
+  it('adds a dynamic layer and rebuilds on updates', function __(done) {
+    const stub = new Stub({}, {
+      stubby: 'property!'
+    });
+
+    properties.dynamic(stub);
+
+    properties.once('build', (props) => {
```
jmanero-r7: Need to initialize the properties instance to register for the layer's update event

test/properties.js
```diff
@@ -0,0 +1,237 @@
...
+  it('activates a view correctly when a source is NO_EXIST', function __(done) {
+    const view = properties.view();
+    const stub = new NoStub();
+
+    // Get current active view's sources
+    const sources = properties.active.sources;
+
+    // Register sources with new view
+    sources.push(stub);
+    sources.forEach((source) => view.register(source));
```
jmanero-r7: TODO: use the constructor interface to add all of the views at once

PR #154, opened 2016-06-03
Fix: No colored logs

fmitchell-r7:
This removes colors from logs. We're logging to STDOUT and letting users handle sending logs to terminals, files, syslog, whatever. Having colors on assumes the output is a color compatible terminal, which may not be the case.


rvanburen-r7:
👍

dgreene-r7:
:shipit:

PR #155, opened 2016-06-06
Fix: Startable test servers

fmitchell-r7:
This allows the local test metadata and S3 servers to start. The S3 server now sets up the same configuration and logging as the metadata server.


jmanero-r7:
👍

PR #156, opened 2016-06-06
Fix: Core API tests

fmitchell-r7:
This fixes up the Core API so all its tests pass. The code has been updated so the output for `/v1/health` and `/v1/status` doesn't change.

### DIFF
--- a/lib/control/v1/core.js
+++ b/lib/control/v1/core.js
@@ -2,6 +2,7 @@
 
 const VERSION = require('../../../package').version;
 const STATUS_CODES = require('../../util/status-codes');
+const Source = require('../../source/common');
 
 const started = Date.now();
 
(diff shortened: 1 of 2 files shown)


jmanero-r7:
Is anything actually relying on these interfaces? These changes remove a lot of important debugging visibility

fmitchell-r7:
@jmanero-r7 We can add more data to the existing APIs. I just don't want to remove anything that's already there without going to `/v2`. Putting `indices` and `sources` in the `/v1/status` API makes sense. Anything additional you want in `/v1/health`?

fmitchell-r7:
We could also just do net new `/v2/health` and `/v2/status` APIs with all the extra debugging information in them.

jmanero-r7:
🚢  it

PR #158, opened 2016-06-06
Refactors Consul source to use Source/Common interface

jmanero-r7:
* initialize() returns a promise
* The Source now only watches `/v1/health/state/any` and derives service health by correlating checks to nodes and services.
* Tests use obfuscated data from a real consul cluster


fmitchell-r7:
If I'm reading [the `update` function](https://github.com/rapid7/propsd/pull/158/files#diff-6b5d4855ba7570774b3223bad74fcacdR63) correctly, if any service on a node is failing, the entire node is in a failing state, right?

jmanero-r7:
```
+      if (serviceName) {
 +        // Set a service-check's status. One-or-more failing check -> Not Passing
 +        if (!services.hasOwnProperty(serviceName)) { services[serviceName] = true; }
 +        services[serviceName] = (check.Status === 'passing') && services[serviceName];
 +      } else if (check.Status !== 'passing') {
 +        // Update the node's global status for a node-check
 +        properties.nodes[nodeID].passing = false;
 +      }
```

If the check has an associated service, its status applies to that service. If it has no associated service (e.g. SerfHealth), it's status applies to all services.

lib/source/consul.js
```diff
@@ -1,358 +1,91 @@
...
+   * Update the catalog cache before parsing results
    *
-   * @param {Error} err
-   * @private
+   * @param  {Object} data Payload from watcher event
    */
-  _error(err) {
-    this._okay = false;
+  _update(data) {
+    this.client.catalog.node.list((err, nodes) => {
```
fmitchell-r7: Do we want to set [the `stale` consistency mode](https://www.consul.io/docs/agent/http.html) when making the call to `/v1/catalog/nodes`?

jmanero-r7: That's a good question. I'm leaning towards no because we're already cache all of the data from Consul in the parser, Rather than propagate flap during an unavailability event on the Consul cluster, we should just wait it out with the existing data.

fmitchell-r7:
👍 Thanks for getting this all working! :shipit:

PR #159, opened 2016-06-07
Fix: Code coverage

fmitchell-r7:
This fixes up code coverage so that `npm run coverage` works again.


fmitchell-r7:
This was resolved in #162.

PR #160, opened 2016-06-07
Update source/common for Consul

jmanero-r7:
* Refactors into a base class and a polling class
* Exports factory methods to bind parsers to source classes
* Implements and tests correct inheritance checking
* Centralizes tests of the Source state-machine

### DIFF
--- a/lib/source/common.js
+++ b/lib/source/common.js
@@ -115,6 +115,7 @@ class Source extends EventEmitter {
   _update(data) {
     // No changes to the source's data have occurred since the last update
     if (data === Source.NO_UPDATE) {
+      this.emit('noupdate', this);
       return this;
     }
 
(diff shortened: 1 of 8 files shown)


fmitchell-r7:
👍 Nice job with the stubs for the tests. It makes the test readable. :shipit:

jmanero-r7:
Thanks. Ok to merge, or do you need more time to review?

dgreene-r7:
Yeah that's elegant as hell. Nice job. Just gotta fix the linting errors and it'll be good to go.

Also, can you rename the PR just for reference?

jmanero-r7:
The lint failures are only in the metadata test module. They're fixed in another commit already

dgreene-r7:
> The lint failures are only in the metadata test module. They're fixed in another commit already

Cool. My one suggestion would be to split the `Source` class and `Polling` class into separate modules but that's really minor.

PR #162, opened 2016-06-07
Refactor some parts of the metadata source for better testability

jmanero-r7:
* Abstract the traverse method into a static utility method
* Use some static obfuscated data from the EC2 metadata API to stub parsing tests
* Remove test cases that have been refactored into the common source module


fmitchell-r7:
Yay, tests!

Is there still a way to [run the metadata server](https://github.com/rapid7/propsd/blob/v1.2.5/package.json#L16-L18) for local testing? In v1.2.5 I can do
```bash
npm run metadata-server &
npm run s3-server -- -d test/data/s3 &
npm start -- -c config/dev.json
```
and have a local instance of propsd to develop against.

lib/source/metadata.js
```diff
@@ -53,90 +50,55 @@ class Metadata extends Source.Polling(Parser) { // eslint-disable-line new-cap
    * @returns {string}
    * @private
```
fmitchell-r7: Since the function name lost its underscore, I'm assuming it's now public. Do we want to drop the `@private` status from the doc comment?

jmanero-r7:
@fmitchell-r7 90becf1 adds a script to serve the mocked metadata paths

fmitchell-r7:
👍 :shipit:

PR #164, opened 2016-06-08
Fix: Consul polling

fmitchell-r7:
This replaces the Consul source with a version based on polling instead of watching. This is equivalent to [the Consul source from v1.2.5](https://github.com/rapid7/propsd/blob/v1.2.5/lib/source/consul.js). We call the `/v1/catalog/services` API periodically to get a list of services, then call the `/v1/health/service/<service>` API to get a list of addresses for healthy services. Consul is polled on the same schedule as we poll S3.


jmanero-r7:
👍 Looks good

PR #165, opened 2016-06-09
Wrap interpolation of source configuration objects during Index creation

jmanero-r7:


fmitchell-r7:
👍 Thanks for the quick fix! :shipit:

PR #166, opened 2016-06-09
Fix: Remove EC2 metadata output

fmitchell-r7:
This removes the EC2 metadata and the static configuration properties from the rendered output. This matches the behavior in v1.2.5 where we used those properties internally but didn't render them.


fmitchell-r7:
This will need to be rebased once #168 is merged.

fmitchell-r7:
Talked with @jmanero-r7 about this. We're going to treat `instance`, `consul`, and `conqueso` as reserved keys in properties. That allows us to delete the EC2 metadata from the output before it's flattened in the `/v1/conqueso` API.

PR #170, opened 2016-06-09
Add an event to update that always fires before the function returns

jmanero-r7:


fmitchell-r7:
Thanks for fixing this! My test case passes now.

1. Run `npm start -- -c config/dev.json`. Wait for the ec2-metadata and s3-index sources to show ERROR messages.
2. Run `npm run metadata-server`. Wait for the ec2-metadata source to become RUNNING.
3. Run `npm run s3-server -- -d test/data/s3`. Wait for the s3-index source to become RUNNING.
4. Check for properties and see that they're empty. `curl -s -XGET http://localhost:9100/v1/conqueso`

👍 :shipit:

PR #171, opened 2016-06-09
Fix: Delete EC2 metadata from Conqueso API

fmitchell-r7:
This reimplements the proposal from #166. We now treat "instance" as a reserved key and simply delete it from the Conqueso API. Documentation is updated to reflect the list of reserved keys.


jmanero-r7:
Thanks @fmitchell-r7. That looks 👍

PR #172, opened 2016-06-09
Fix: Allows layers with undefined properties

fmitchell-r7:
This fixes a bug where propsd wouldn't start if you didn't have any properties defined statically in your config file.

### DIFF
--- a/lib/properties.js
+++ b/lib/properties.js
@@ -148,6 +148,11 @@ class Properties extends EventEmitter {
    * @return {Object}               The destination object
    */
   static merge(destination, source) {
+    // Only merge source if it's an Object. `== null ` covers both null and undefined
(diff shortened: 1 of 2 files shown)


lib/properties.js
```diff
@@ -148,6 +148,11 @@ class Properties extends EventEmitter {
    * @return {Object}               The destination object
    */
   static merge(destination, source) {
+    // Only merge source if it's an Object. `== null ` covers both null and undefined
+    if (source == null || Object.getPrototypeOf(source) !== Object.prototype) {
+      return destination;
+    }
+
```
jmanero-r7: This should go after the `destination` check so that Properties.merge always returns an Object

jmanero-r7:
👍

PR #174, opened 2016-06-10
Feature: Document Consul config

### LINKED ISSUE #138: Document how the Consul plugin works
We're missing documentation in the [getting started guide](https://github.com/rapid7/propsd/tree/v1.2.2/docs/getting-started) around the [Consul plugin](https://github.com/rapid7/propsd/blob/v1.2.2/lib/source/consul.js). We should include a section that describes how propsd can be configured to leverage [Consul](https://www.consul.io/) for service discovery. This section should also describe how propsd returns discovered services to keep it compatible with [Conqueso](https://github.com/rapid7/conqueso) and the conditions under which services will be discovered e.g. all their Consul health checks are in a passing state.

fmitchell-r7:
This fixes #138. Documentation around how to configure propsd so it can use [Consul](https://www.consul.io/) for service discovery is provided. Additionally, the example `/v1/health` and `/v1/status` outputs have been updated to reflect the new metadata included in this refactor.

### DIFF
--- a/docs/getting-started/configuration.md
+++ b/docs/getting-started/configuration.md
@@ -94,6 +94,24 @@ The configuration file below is the default settings for propsd.
   * `interval` - The time in milliseconds to poll the index property file for
     changes. Defaults to 30000 (30 seconds).
 
+* `consul` - These settings control service discovery via [Consul][].
+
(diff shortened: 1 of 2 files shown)


jmanero-r7:
👍

PR #176, opened 2016-06-13
Fix: Remove colored output from request logs

### LINKED ISSUE #173: Request logs in have colored messages
I'm seeing colored messages in request logs in v1.2.5.
```json
{ message: "\u001b[90mGET /v1/conqueso\u001b[39m \u001b[32m200\u001b[39m \u001b[90m2ms\u001b[39m" }
```
These appear to be coming from [express-winston](https://github.com/bithavoc/express-winston/blob/v1.3.0/index.js#L280-L286). We can fix that by explicitly defining our own `msg` attribute instead of enabling the `expressFormat` attribute when [configuring logging](https://github.com/rapid7/propsd/blob/v1.2.5/lib/logger.js#L63).
```javascript
msg: "{{req.method}} {{req.url}} {{res.statusCode}} {{res.responseTime}}ms"
```

fmitchell-r7:
This fixes #173 and removes colored output from the request logs in propsd.

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -61,7 +61,7 @@ function RequestLogger(logger, level) {
   return expressWinston.logger({
     winstonInstance: logger,
     expressFormat: false,
-    msg: "{{req.method}} {{req.url}} {{res.statusCode}} {{res.responseTime}}ms",
+    msg: '{{req.method}} {{req.url}} {{res.statusCode}} {{res.responseTime}}ms',
     level: logLevel,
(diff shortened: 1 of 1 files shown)


dgreene-r7:
Other than the linter error, looks good!

fmitchell-r7:
@dgreene-r7, I fixed the lint error in https://github.com/rapid7/propsd/pull/176/commits/a3bdf94064412aa64130efbfa12dfab9ee069171.

PR #177, opened 2016-06-23
This changeset reduces the number and rate of requests that the ec2-metadata provider makes of the ec2 metadata service

jmanero-r7:
* Reduce parallel requests to 1 (no parallelism)
* Drive an explicit enumeration of the paths to fetch from the metadata service from the parser.

* Log the number of paths retrieved from the metadata service. This may be a useful indicator if empty responses are a cause of issues with the Metadata source


fmitchell-r7:
👍  Thanks for fixing this! :shipit:

rvanburen-r7:
👍

PR #179, opened 2016-07-25
Fix: S3 ETag caching

### LINKED ISSUE #178: ETag isn't being cleared on S3 source when it's not found
When a S3 source errors or returns a `NoSuchKey` message, the ETag isn't cleared. This results in properties not showing up when a S3 source file is removed and then replaced. If the contents of the source are updated (changing the ETag) the properties for the source will show up again.

### Steps to reproduce ###

1. Set up an index file with two S3 sources. The first source should contain a string property with a default empty value. The second source should contain the same string property with a non-empty overwrite value.
2. Stat propsd.
3. Read the `/v1/conqueso` end point. The overwrite value will be returned.
4. Remove the second source file from S3.
5. Read the `/v1/conqueso` end point. The default value will be returned.
6. Replace the second source file in S3.
7. Read the `/v1/conqueso` end point. The default value will be returned.

### Expected behavior ###

Properties from S3 sources are reevaluated when the sources are removed and added back even if their ETags don't change. The above steps should have the non-default value returned at the end.

fmitchell-r7:
This fixes #178. The ETag for an S3 source is now explicitly cleared when a `NoSuchKey` error is returned from AWS. This ensures propsd will get properties from the source when it becomes available.

The fix applies to all polling sources. Currently, only the S3 source takes advantage of it.

### DIFF
--- a/lib/source/common.js
+++ b/lib/source/common.js
@@ -139,7 +139,7 @@ class Source extends EventEmitter {
 
       // Emit an update if the source no longer exists
       if (previousState !== Source.WAITING) {
-        Log.log('DEBUG', `Source ${this.type} source ${this.name} no longer exists`, this.status());
+        Log.log('INFO', `Source ${this.type} source ${this.name} no longer exists`, this.status());
         this.emit('update', this);
       }
 
--- a/lib/source/common.js
+++ b/lib/source/common.js
@@ -105,6 +105,8 @@ class Source extends EventEmitter {
     return this.state !== Source.ERROR && this.state !== Source.WARNING;
   }
 
+  /* eslint-disable max-statements */
+
   /**
    * Called by implementations to update source data
    *
@@ -169,6 +171,8 @@ class Source extends EventEmitter {
     this.emit('update', this);
     return this;
   }
+
+  /* eslint-enable max-statements */
 
   /**
    * Handle errors from underlying source facilities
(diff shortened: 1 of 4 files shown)


jmanero-r7:
After reviewing the code for the _update method more closely, I think we need to collapse [that switch case](https://github.com/rapid7/propsd/blob/f744fc3db7fdbabc9a52a9491c017c19fd176d61/lib/source/common.js#L124-L156) into a much simpler condition:


```
_update(...) {
  const previousState = this.state;
  
 if (data === Source.NO_UPDATE) {
    this.emit('noupdate', this);
    return this;
  }

  if (data === Source.NO_EXIST) {
    this.state = Source.WAITING;

    if (previousState == Source.WAITING) {
      return this;
    }

    // Clear previous properties and state
    this.properties = {};
    this._state = null;
    this.updated = new Date();

    if (previousState === Source.INITIALIZING) {
      this.emit('initialized', this);
    }

    Log.log('DEBUG', `Source ${this.type} source ${this.name} no longer exists`, this.status());

    // Notify watchers
    this.emit('update', this);
    return this;
  }
}
```

fmitchell-r7:
@jmanero-r7, should we always be resolving the `initialize()` Promise after setting `properties` and `sources`? We're [not doing that](https://github.com/rapid7/propsd/blob/v1.2.7/lib/source/common.js#L170-L179) for the case where the data has actually changed.

jmanero-r7:
Only `if (previousState === Source.INITIALIZING)`

jmanero-r7:
👍

PR #181, opened 2016-08-30
Make cookbook version-independent

dgreene-r7:
This PR changes the packaging target to install propsd into `propsd-VERSION` so we can install multiple versions. It also symlinks the current version to `node['propsd']['paths']['directory']` (nominally `/opt/propsd`) and `chowns` the directory to the propsd user.


fmitchell-r7:
👍 Thanks for getting this working!

Do we want to do a minor bump on the version? Technically, we're just changing the packaging, not the API, but it feels like we should cut a new release for this. Thoughts?

dgreene-r7:
Yeah, I think it deserves a minor. It changes one of propsd's APIs (packaging & base cookbook) in such a way that it might break changes in upstream cookbooks and the release should reflect that.

PR #186, opened 2016-09-06
Add Propsd Ohai Plugin

athompson-r7:
This PR adds an Ohai plugin for Propsd.  This gives you access to all of the properties found in the Propsd `/properties` api from within Chef.

This plugin is optional, in-order to use it add `propsd::ohai_plugin` to a run_list.

The of the properties are available under the `node['propsd_plugin']` key.

### DIFF
--- a/Buildfile
+++ b/Buildfile
@@ -7,6 +7,6 @@ cookbook.depends 'propsd' do |propsd|
   propsd.path './cookbook'
 end
 
-profile :default do |default|
-  default.chef.run_list 'propsd::default'
+profile :default do |test|
+  test.chef.run_list 'propsd::test'
 end
--- a/cookbook/files/default/propsd_ohai_plugin.rb
+++ b/cookbook/files/default/propsd_ohai_plugin.rb
@@ -1,0 +1,70 @@
+require 'json'
+
+Ohai.plugin(:Propsd) do
+  provides 'propsd_plugin'
+
+  PROPSD_HOST = 'localhost' unless defined?(PROPSD_HOST)
+  PROPSD_PORT = 9100 unless defined?(PROPSD_PORT)
+
+  def can_propsd_connect?(addr, port, timeout = 2)
+    t = Socket.new(Socket::Constants::AF_INET, Socket::Constants::SOCK_STREAM, 0)
+    saddr = Socket.pack_sockaddr_in(port, addr)
+    connected = false
+
+    begin
+      t.connect_nonblock(saddr)
+    rescue Errno::EINPROGRESS
+      r, w, e = IO.select(nil, [t], nil, timeout)
+      if !w.nil?
+        connected = true
+      else
+        begin
+          t.connect_nonblock(saddr)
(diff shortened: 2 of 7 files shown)


cookbook/files/default/propsd_ohai_plugin.rb
```diff
@@ -0,0 +1,70 @@
...
+        rescue SystemCallError
+        end
+      end
+    rescue SystemCallError
+    end
+    Ohai::Log.debug("can_propsd_connect? == #{connected}")
+    connected
+  end
+
+  def get_properties_from_file
```
fmitchell-r7: Is this method ever called? I can't find a reference to it.

cookbook/recipes/ohai_plugin.rb
```diff
@@ -0,0 +1,13 @@
...
+# Recipe:: ohai_plugin
+#
+# Copyright (C) 2016 Rapid7 LLC.
+#
+# Distributed under terms of the MIT License. All rights not explicitly granted
+# in the MIT license are reserved. See the included LICENSE file for more details.
+#
+
+ohai_plugin 'propsd_ohai_plugin' do
+  path node['propsd']['ohai_plugin_path']
```
fmitchell-r7: It looks like the default value for `node['propsd']['ohai_plugin_path']` is `nil`. Where does the plugin get installed in that case?

Buildfile
```diff
@@ -7,6 +7,6 @@ cookbook.depends 'propsd' do |propsd|
   propsd.path './cookbook'
 end
 
-profile :default do |default|
-  default.chef.run_list 'propsd::default'
+profile :default do |test|
```
dgreene-r7: Instead of doing this you should add a new `:test` profile that extends `:default`. For testing you can do `build local test` and the default profile is preserved.

cookbook/recipes/ohai_plugin.rb
athompson-r7: > Where does the plugin get installed in that case?

It get that value from the ohai cookbook, which ends up being `::File.join(chef_config_path, 'ohai', 'plugins')`.

https://github.com/chef-cookbooks/ohai/blob/master/resources/plugin.rb#L33

cookbook/files/default/propsd_ohai_plugin.rb
athompson-r7: > Is this method ever called? I can't find a reference to it.

Shoot, forgot to remove that.  Was there for testing without Propsd there.

Buildfile
athompson-r7: That's fine but, the default profile doesn't give you an instance that even starts.  What is the default profile for if not for local testing?

dgreene-r7: I still don't love that `default` maps to `test` but you're right that at this point there's no distinction.

fmitchell-r7:
👍  Thanks for putting this together. 🚢  it.

PR #188, opened 2016-09-07
Feature: Parse secrets from Tokend

### LINKED ISSUE #185: Parse generic secrets from tokend
The **Tokend Transformer** is responsible for transforming generic secrets defined in property files by retrieving those secrets from [tokend](https://github.com/rapid7/tokend). See #184 for the definition of a generic secret within a property file.

An instance of the `TokendTransformer` class has the following methods:

**constructor(options)**
Create a new instance of the `TokendTransformer`.

The `options` object may contain "host" and "port" keys defining connection end points for tokend. If those keys are not provided, tokend's default settings should be used.

**transform(properties)**
Transforms properties by retrieving secrets from tokend. Returns a `Promise` that resolves with the transformed secrets.

If no secrets where found in `properties`, and no errors encountered while parsing, the promise resolves with an empty object. If parsing failed or secrets could not be retrieved from tokend, the promise resolves with an `Error`.

The original `properties` are treated as immutable and never modified. Merging of transformed secrets into properties should be handled elsewhere. TBD where that is.

Because secrets are stored in Vault as key/value pairs, this implementation can assume secrets are stored in Vault with a key of "plaintext".

fmitchell-r7:
This fixes #185. A `TokendTransformer` class is provided to handle the retrieval and parsing of secrets from Tokend.


package.json
```diff
@@ -59,6 +60,7 @@
     "depd": "^1.1.0",
     "express": "^4.13.3",
     "flat": "^2.0.0",
+    "immutable": "^3.8.1",
```
dgreene-r7: Oh hell yes.

lib/transformers/tokend.js
```diff
@@ -0,0 +1,112 @@
...
+          res.on('data', (chunk) => {
+            data += chunk;
+          });
+          res.on('end', () => {
+            try {
+              data = JSON.parse(data);
+            } catch (e) {
+              return reject(e);
+            }
+            if (data.hasOwnProperty('plaintext')) {
```
dgreene-r7: I assume this means we're going to write Vault secrets as `{plaintext: <VALUE>}`? That's totally ok, just want to make sure once we start loading secrets that we follow that pattern.

fmitchell-r7: Yes. Secrets stored in Vault are in key value format. I thought about exposing the key as part of the configuration...
```json
"$tokend": {
  "type": "generic",
  "key": "plaintext",
  "resource": "/v1/secret/default/kali-root-password"
}
```
I figured it was simpler to go with a known key to get something working. If there's a need, we can expose configurable keys later and fall back to checking for a "plaintext" key if no key's given in the configuration.

dgreene-r7:
As a slight aside, I'm going to add a ticket to refactor some of the codebase to use Immutable.js so we don't run into stuff like the `reverse()` bug again.

lib/transformers/tokend.js
```diff
@@ -0,0 +1,112 @@
...
+
+  // Ensure we're walking an Object. `== null` covers both null and undefined.
+  if (properties == null || Object.getPrototypeOf(properties) !== Object.prototype) { // eslint-disable-line eqeqeq
+    return results;
+  }
+
+  // If we're walking a $tokend object, pass it to the callback.
+  const keys = Object.keys(properties);
+
+  if (keys.length === 1 && keys[0] === '$tokend') {
```
dgreene-r7: I'm just going to make a note that we'll definitely have to revisit this when we add more transformers.

fmitchell-r7:
Talking with @jmanero-r7 about this, and we're thinking this plugs into [Properties#build()](https://github.com/rapid7/propsd/blob/v1.4.1/lib/properties.js#L129) after properties have been merged and before the `build` event is emitted. Once the `$tokend` properties have been transformed, we emit the `build` event and notify listeners that new properties are available.

Putting the transformation means the `TokendTrasnformer` has to only walk the properties object once and can use caching to avoid looking up the same secret multiple times. It also means that changes to `$tokend` objects trigger secrets being reevaluated without any additional logic requirements.

There's an additional complication where the secrets may change in Vault itself. In that case, we want to trigger `Properties#build()` so secrets are reevaluated and recomposed. Handling that case means `TokendTransformer` may need to become a [Polling Source](https://github.com/rapid7/propsd/blob/v1.4.1/lib/source/common.js#L246) so it can periodically check if secrets have changed in Vault.

@dgreene-r7, does this make sense to you?

dgreene-r7:
> @dgreene-r7, does this make sense to you?

This looks completely reasonable. I was thinking this over last week and I came to the same conclusion, that we want any transformations to occur as close after the final set of properties are assembled as we can.

There's no reason for us to walk the properties of each source and grab every secret when we can wait until all of the layering has finished before we go out and get the secrets that will actually be displayed in the final set of properties.

As far as source type, here are my thoughts: we should have every property rebuild trigger a fresh run of the `TokendTransformer` and retrieval of secrets from Vault. If we try to cache secrets across rebuilds and poll Tokend for changes we'll run into all sorts of timing issues and frankly, it's not worth the effort and complexity to prematurely optimize.

fmitchell-r7:
> ...we should have every property rebuild trigger a fresh run of the `TokendTransformer` and retrieval of secrets from Vault.

Yes.

> If we try to cache secrets across rebuilds and poll Tokend for changes we'll run into all sorts of timing issues and frankly, it's not worth the effort and complexity to prematurely optimize.

We need to handle the case where sources haven't changed (no rebuilds are triggered) but secrets have changed in Vault (we want to trigger a rebuild). Polling on the last set of secrets we got after a build seems like it accomplishes this? But maybe there's a simpler solution I'm missing.

dgreene-r7:
Ok, I see what you're getting at. 

Having the `TokendTransformer` emit an event that `Properties` listens for, then using that event to trigger a rebuild whenever tracked secrets change definitely solves the problem. I think it's a fairly simple addition to what you have so far so I say, "go for it!"

fmitchell-r7:
@jmanero-r7 and @dgreene-r7, would you mind taking another look at this, please? In particular pay attention to...

Error handling in [TokendTransformer#transform](https://github.com/rapid7/propsd/pull/188/files#diff-43f75ed32f05b13da5671916c948d3c8R58). Secrets that can't be parsed or retrieved log warnings and resolve with `null` values. Is that the behavior we want?

Secret caching in [TokendClient#get](https://github.com/rapid7/propsd/pull/188/files#diff-0b03fcffe3135918741d9af5b8916634R37). This seems to work, but I'm wondering if there's a better pattern I should be using.

lib/properties.js
```diff
@@ -129,11 +132,14 @@ class Properties extends EventEmitter {
       // Merge layers' properties into the global property object
       Properties.merge(properties, persistent);
 
-      this.persistent = persistent;
-      this.properties = properties;
-      this.emit('build', this.properties);
+      this.tokendTransformer.transform(properties)
+          .then((transformedProperties) => {
+            this.persistent = persistent;
+            this.properties = Immutable.Map(properties).mergeDeep(transformedProperties).toJS();
```
fmitchell-r7: This is intentional. Our `Properties.merge` function won't overwrite existing values with `null` values. We're using `null` to indicate that a secret wasn't found. `Immutable.mergeDeep` will overwrite the existing `$tokend` objects in the properties with either resolved secrets or `null` values.

fmitchell-r7:
@jmanero-r7, I was running into issues updating properties in place. Using `Immutable`, and it's [Map.setIn](https://facebook.github.io/immutable-js/docs/#/Map/setIn) function in particular, got me around that.

lib/transformers/tokend-client.js
```diff
@@ -0,0 +1,70 @@
...
+          });
+        });
+
+        req.on('error', (e) => {
+          reject(e);
+        });
+      });
+    }
+
+    return Promise.resolve(this._pendingGetRequests[path]);
```
fmitchell-r7: Thanks for spotting that! It's fixed in ee4eb529345204fce3a76ec3273fd7f83eff9425.

fmitchell-r7:
@jmanero-r7 and @dgreene-r7, I've got the update functionality working, so when secrets/keys in Vault change, we'll rebuild properties. Can you take another look, please?

PR #191, opened 2016-09-08
Fixes Mash name

athompson-r7:
This change fixes the name of the Mash we are assigning values.  This
was missed in a final refactor to #186.


fmitchell-r7:
👍 Thanks for fixing this! 🚢 it.

PR #192, opened 2016-09-08
Bug layer ordering gets reversed

### LINKED ISSUE #189: Bug: Layer ordering can end up in a reversed state

dgreene-r7:
This PR refactors parts of the `Properties` and `View` classes to use immutable objects as data storage rather than plain JS objects. This fixes problems where certain built in methods were mutating objects in place.

It also pins the node version to the latest LTS/argon release and reorders the way the Consul, Metadata, and local Config layers are loaded.

Resolves #189.


fmitchell-r7:
Thanks for fixing this!

Would you mind adding some tests in to make sure this bug stays fixed? Specifically, I'm thinking there should be tests for a) ensuring multiple calls to `Properties.build()` don't reverse the layer ordering, and b) ensuring calls to the `/v1/health` and `/v1/status` APIs don't reverse the active sources ordering.

fmitchell-r7:
👍 Thanks for adding those tests! I'd get a 👍 from @jmanero-r7 as well to make sure there's nothing subtle we're missing here, then 🚢 it.

rvanburen-r7:
👍

lib/properties.js
```diff
@@ -13,8 +14,8 @@ class Properties extends EventEmitter {
 
     this.initialized = false;
 
-    this.layers = [];
-    this.properties = {};
+    this._layers = Immutable.List();
+    this._properties = Immutable.Map();
```
jmanero-r7: I don't think that introducing additional complexity is the right way to solve this. Rather, we should refactor the parts that actually (unnecessarily) require usage of `reverse` in the first place, making the model simpler and more consistent with the interfaces it's supposed to present.

dgreene-r7: `Immutable` is going to be introduced in #188 anyway. Not sure why we shouldn't take advantage of it.

dgreene-r7:
Superseded by #196.

PR #193, opened 2016-09-12
Add a feature flag attribute to disable service and notifications

jmanero-r7:


rvanburen-r7:
👍

fmitchell-r7:
I noticed the "description" is empty. Is this just "let's make testing in Vagrant easier" or is there something deeper going on here (something broken) that this fixes?

dgreene-r7:
I'd add a description to the PR like, @fmitchell-r7 said. Then :shipit:

PR #195, opened 2016-09-15
Suppress EC2 metadata client errors for non-existent paths

jmanero-r7:
Somewhere in aws-sdk@2.6 the behavior of the metadata service client changed to start raising errors for 404 responses. This change catches and suppresses those errors.

### DIFF
--- a/lib/source/metadata.js
+++ b/lib/source/metadata.js
@@ -64,16 +64,16 @@ class Metadata extends Source.Polling(Parser) { // eslint-disable-line new-cap
 
       // Call `Metadata.request` for each path
       (path, cb) => this.service.request(path, (err, value) => {
-        if (err && err.message === null) {
+        if (err) {
           /*
-           * This is really great. AWS-SDK > 2.6ish now raises an error with a null
-           * message when a requested metadata path doesn't exist. We don't want
-           * to abort the rest of the traversal for this.
+           * AWS-SDK > 2.6.0 now raises an error with a null message when the underlying http
+           * request returns a non-2xx status code. We don't want to abort the rest of the traversal
+           * for this. Instead, log the error and swallow it.
            */
-
-          return cb(null, false);
+          Log.log('ERROR', 'Aws-sdk returned the following error during the metadata service request ' +
+              `to ${path}: %j`, err);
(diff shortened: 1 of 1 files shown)


dgreene-r7:
This is a solid solution but it makes assumptions that upstream the sdk won't change the `Error` that's being raised.

Let's consider the possible error conditions here:

1. The endpoint is missing (404)
2. There's a server error in the metadata service (500)
3. We're rate limited

The sdk now includes a retry mechanism that should get us around seeing an error for rate limiting, based on what we're doing. 

For http errors, I submit that we should swallow both of those errors and log their occurrence. Obviously without more information from the sdk we won't know what kind of error we get, but having some kind of info when we get an error back from metadata requests would be helpful.

So, IMO, a better solution looks more like this:
```javascript
(path, cb) => this.service.request(path, (err, value) => {
  if (err) {
    /*
     * AWS-SDK > 2.6.0 now raises an error with a null message when the underlying http
     * request returns a non-2xx status code. We don't want to abort the rest of the traversal 
     * for this. Instead, log the error and swallow it.
     */

    Log.log('ERROR', `Aws-sdk returned the following error during the metadata service request to ${path}: %j`, err);
    cb(null, false);
  }
  cb(err, value);
}),
```

lib/source/metadata.js
```diff
@@ -63,7 +63,19 @@ class Metadata extends Source.Polling(Parser) { // eslint-disable-line new-cap
...
+      (path, cb) => this.service.request(path, (err, value) => {
+        if (err) {
+          /*
+           * AWS-SDK > 2.6.0 now raises an error with a null message when the underlying http
+           * request returns a non-2xx status code. We don't want to abort the rest of the traversal
+           * for this. Instead, log the error and swallow it.
+           */
+          Log.log('ERROR', 'Aws-sdk returned the following error during the metadata service request ' +
+              `to ${path}: %j`, err);
+          cb(null, false);
```
fmitchell-r7: Do you need an early return here so the second callback doesn't fire?
```javascript
return cb(null, false);
```

dgreene-r7: Yup! Good catch.

PR #196, opened 2016-09-16
Bug: Reversed layer ordering

### LINKED ISSUE #189: Bug: Layer ordering can end up in a reversed state

dgreene-r7:
This PR refactors the `Properties` class to never reverse internal arrays. This fixes problems where the `Properties#sources` getter and `Properties#build()` were mutating objects in place. It reverses the order in which Consul, EC2 Metadata, and static properties are loaded in `server.js`.

It also pins the node version to the latest LTS/argon release and reorders the way the Consul, Metadata, and local Config layers are loaded.

Resolves #189 and supersedes #192.


fmitchell-r7:
@dgreene-r7, would you mind splitting this into two PRs, one for the bug fix with tests and one for the eslint and doc improvements? I feel like it makes writing the changelog for releases easier when I can say, "Here's a PR that fixes a bug" and "Here's a PR that improves documentation".

dgreene-r7:
Can do!

lib/properties.js
```diff
@@ -29,9 +32,8 @@ class Properties extends EventEmitter {
       // Get Dynamic layers' source instances
       .concat(this.layers.map((layer) => layer.source).filter((source) => !!source))
 
-      // Add the active View's sources. Reversed for precedence consistency with Layers:
-      // Entry 0 is most precedent, LAST is least precedent.
-      .concat(this.active.sources.reverse());
+      // Add the active View's sources.
+      .concat(this.active.sources);
```
jmanero-r7: I had to stare at this for a while to make sure...

.nvmrc
```diff
@@ -1 +1 @@
-4.4.0
+lts/argon
```
dgreene-r7: Crap, you're right. I'm splitting the stuff out into 2 separate PRs so I'll drop this change.

lib/properties.js
dgreene-r7: I feel like this is diminishing returns. The only use of this getter is in the health check and the order doesn't change anything beyond which order the source statuses are displayed in the `/v1/status` output.

jmanero-r7: Yep. _shrugs_ Keeping the API unchanged is basically what I'm getting at.

PR #197, opened 2016-09-16
Eslint and doc improvements

dgreene-r7:
This PR is half of the code submitted in #196. It surrounds centralizing `eslint` settings for tests and adding jsdocs where they're missing.


fmitchell-r7:
👍  Thanks for cleaning this up! 🚢  it.

PR #198, opened 2016-09-16
Bug layer ordering reversed

### LINKED ISSUE #189: Bug: Layer ordering can end up in a reversed state

dgreene-r7:
This PR refactors the `Properties` class to never reverse internal arrays. This fixes problems where the `Properties#sources` getter and `Properties#build()` were mutating objects in place. It reverses the order in which Consul, EC2 Metadata, and static properties are loaded in `server.js`.

Resolves #189 and supersedes #192 and #196.

### DIFF
--- a/lib/properties.js
+++ b/lib/properties.js
@@ -29,9 +29,8 @@ class Properties extends EventEmitter {
       // Get Dynamic layers' source instances
       .concat(this.layers.map((layer) => layer.source).filter((source) => !!source))
 
-      // Add the active View's sources. Reversed for precedence consistency with Layers:
-      // Entry 0 is most precedent, LAST is least precedent.
-      .concat(this.active.sources.reverse());
+      // Add the active View's sources.
+      .concat(this.active.sources);
   }
 
   /**
--- a/lib/properties.js
+++ b/lib/properties.js
@@ -112,9 +112,6 @@ class Properties extends EventEmitter {
       const persistent = {};
 
       this.layers
-
-        // Merge layers in reverse. `layers[0]` is the most precedent
-        .reverse()
         .forEach((layer) => {
           if (!layer.namespace) {
             return Properties.merge(persistent, layer.properties);
--- a/lib/properties.js
+++ b/lib/properties.js
@@ -30,6 +30,8 @@ class Properties extends EventEmitter {
       .concat(this.layers.map((layer) => layer.source).filter((source) => !!source))
 
       // Add the active View's sources.
+      //  NOTE: This is a shallow copy and only copies object references to a new array.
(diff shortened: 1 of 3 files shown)


lib/properties.js
```diff
@@ -29,9 +29,8 @@ class Properties extends EventEmitter {
       // Get Dynamic layers' source instances
       .concat(this.layers.map((layer) => layer.source).filter((source) => !!source))
 
-      // Add the active View's sources. Reversed for precedence consistency with Layers:
-      // Entry 0 is most precedent, LAST is least precedent.
-      .concat(this.active.sources.reverse());
+      // Add the active View's sources.
+      .concat(this.active.sources);
```
fmitchell-r7: Do we want to keep the API unchanged for `/v1/status` and have this go out as a bug fix? If so, we can write this as
```javascript
return []
  .concat(this.active.sources)
  .concat(this.layers.map((layer) => layer.source).filter((source) => !!source))
  .reverse();
```

dgreene-r7: If we want to keep it unchanged we'd do:
```javascript
return []
  .concat(this.layers.map((layer) => layer.source).filter((source) => !!source))
  .concat(this.active.sources.slice().reverse());
```

Keep in mind the reversal was applying to the order in which active sources were listed in the `/v1/status` endpoint so we have to reverse `this.active.sources` before we `concat` it with the rest of the sources.

I'm ambivalent. Let me know your preference and we can do it however you want.

fmitchell-r7: Let's keep the API unchanged and have this go out as a bug fix.

test/properties.js
```diff
@@ -34,6 +34,55 @@ describe('Properties', function _() {
...
+
+  it('does not reorder layers if the properties sources getter is called', function ___(done) {
+    const props = new Properties();
+    const sources = [
+      new Source.Stub({path: 'foo'}),
+      new Source.Stub({path: 'bar'}),
+      new Source.Stub({path: 'baz'})
+    ];
+    const view = props.view(sources);
+    const expected = ['baz', 'bar', 'foo'];
```
fmitchell-r7: I can't get this test to fail correctly against the v1.4.2 code line. Try this instead?
```javascript
const expected = ['foo', 'bar', 'baz'];
const reversed = ['baz', 'bar', 'foo'];

props.once('build', () => {
  expect(props.sources.map((s) => s.properties.path)).to.eql(reversed);
  expect(props.active.sources.map((s) => s.properties.path)).to.eql(expected);
  done();
});

view.activate();
```

The value returned from `props.sources` _should_ be reversed, but the value returned from `props.active.sources` _should not_ be reversed.

fmitchell-r7:
👍  Thanks for fixing this! 🚢  it.

PR #199, opened 2016-10-03
GitHub release task

dgreene-r7:
This PR repurposes the `:release` task to create a new Github release. Running this task requires proper repository permissions or else the task will fail.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -40,11 +40,11 @@ def repo
 end
 
 def target_version
-  ::File.read(::File.join(@base_dir, '.nvmrc')).strip()
+  ::File.read(::File.join(base_dir, '.nvmrc')).strip.delete('v')
 end
 
 def max_version
-   target_version.split('.').first.to_f + 1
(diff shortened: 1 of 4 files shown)


Rakefile
```diff
@@ -69,6 +91,31 @@ task :release => [:install, :shrinkwrap, :pack] do
...
+  rescue Octokit::NotFound
+    latest_release = OpenStruct.new(name: 'master')
+  end
+
+  release = github_client.create_release(
+    github_repo,
+    "v#{version}",
+    :name => "v#{version}",
+    :draft => true
+  )
```
athompson-r7: Will this fail if the release already exists?  Such as, in the event the version was not bumped.

athompson-r7:
This is awesome @dgreene-r7

Rakefile
dgreene-r7: I imagine so and imo, it should fail and fail loudly.

PR #212, opened 2016-10-26
Feature: Transit secret provider

### LINKED ISSUE #211: Parse transit secrets from Tokend
Now that [Tokend supports decrypting transit secrets](https://github.com/rapid7/tokend/issues/59), we want to be able to use them in Propsd files. We can extend the [TokendTransformer class](https://github.com/rapid7/propsd/blob/master/lib/transformers/tokend.js) to parse transit secrets in addition to the generic secrets it handles now. We'll also need to update the [TokendClient class](https://github.com/rapid7/propsd/blob/master/lib/transformers/tokend-client.js) so it can track POST requests.

fmitchell-r7:
This is a work in progress to fix #211 and allow us to parse transit secrets decrypted by Tokend.

### DIFF
--- a/lib/properties.js
+++ b/lib/properties.js
@@ -83,18 +83,14 @@ class Properties extends EventEmitter {
     this.initialized = true;
 
     return Promise.all(
-      this.layers.map((layer) => layer.initialize())
-    ).then((layers) => {
-      // Once initialized, watch for layers' update events
-      layers.forEach((layer) => {
-        layer.on('update', () => this.build());
+      this.layers.concat(this.tokendTransformer).map((source) => source.initialize())
+    ).then((sources) => {
+      // Once initialized, watch for sources' update events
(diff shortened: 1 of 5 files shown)


jmanero-r7:
The way these tests are failing kind of looks like a race condition in the triggering logic

image

fmitchell-r7:
@jmanero-r7, yes. The goal of that test is to ensure that when a generic secret changes in Vault, we pick up the change and propagate it to the properties.

So we've got a polling loop in the TokendClient to watch for changes to generic secrets in Vault. That emits an `update` event which triggers `build()`, but `build()` may or may not actually do anything at that time because it has a hold down timer.

Currently I'm trying to mock a polling event by restarting the TokendClient.
```javascript
_properties.tokendTransformer._client.stop().start();
```
But it really feels like this test is pretty brittle and relies too heavily on timing.

PR #213, opened 2016-11-08
Adding support for Ubuntu 16.04 via systemd service support

cdoughty-r7:
Currently this cookbook does not support systems that depend on systemd to start services.  

This change adds support for systemd.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,4 +8,8 @@ gem 'mime-types', '~> 3.1'
 
 group :cookbook do
   gem 'builderator', '~> 1.0'
+  gem 'test-kitchen'
+  gem 'kitchen-docker'
+  gem 'kitchen-vagrant'
+  gem 'berkshelf', '~> 4.3'
 end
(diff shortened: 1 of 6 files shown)


cdoughty-r7:
@fmitchell-r7 @dgreene-r7 Ready for review, if either of you get a chance.

Gemfile
```diff
@@ -8,4 +8,8 @@ gem 'mime-types', '~> 3.1'
 
 group :cookbook do
   gem 'builderator', '~> 1.0'
+  gem 'test-kitchen'
+  gem 'kitchen-docker'
```
erran-r7: Why include the docker driver if we don't use it?

cdoughty-r7: copy/paste.  I'll remove.

Gemfile.lock
```diff
@@ -136,6 +137,10 @@ GEM
     ipaddress (0.8.3)
     jmespath (1.3.1)
     json (2.0.2)
+    kitchen-docker (2.6.0)
```
fmitchell-r7: @cdoughty-r7, can you regenerate the Gemfile.lock so `kitchen-docker` is removed from it as well?

cdoughty-r7:
@fmitchell-r7 Added, thanks for catching that.

PR #228, opened 2017-03-07
Feature: retrieve specific property

### LINKED ISSUE #227: Properties endpoint should allow specific key retrieval
In order to support https://github.com/rapid7/turnstile/issues/30 the `/v1/properties/*` endpoint should allow retrieving a single key. It should gracefully handle nested keys that don't exist.

dgreene-r7:
This PR adds an additional `/v1/properties/:property*` endpoint to allow specifying a single property for retrieval.

This resolves #227.

### DIFF
--- a/lib/control/v1/properties.js
+++ b/lib/control/v1/properties.js
@@ -20,16 +20,9 @@ exports.attach = function attach(app, storage) {
       return next(new Error(`Property ${prop} not found`));
     }
 
-    const resp = {};
     const value = properties[prop];
     const extra = req.params[0].split('/').filter(Boolean);
 
-    if (typeof value === 'object' && extra.length === 0) {
-      resp[prop] = value;
-      return res.json(resp);
-    }
-
-    // iterate `extra` until it's empty and we have the property we want
     res.json(getNestedProperty(value, Array.from(extra)));
   });
 
--- a/lib/control/v1/properties.js
+++ b/lib/control/v1/properties.js
@@ -30,8 +30,7 @@ exports.attach = function attach(app, storage) {
     }
 
     // iterate `extra` until it's empty and we have the property we want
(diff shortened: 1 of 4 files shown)


lib/control/v1/properties.js
```diff
@@ -1,13 +1,37 @@
...
+      return next(new Error(`Property ${prop} not found`));
+    }
+
+    const value = properties[prop];
+    const extra = req.params[0].split('/').filter(Boolean);
+
+    res.json(getNestedProperty(value, Array.from(extra)));
+  });
+
+  app.get('/v1/properties*', function handler(req, res) {
```
erran-r7: I'm guessing the vanilla `GET /v1/properties` (without anything additional) is still supported?

dgreene-r7: Yeah, that's what the `/v1/properties*` route covers, although that `*` glob at the end doesn't really need to be there.

lib/logger.js
```diff
@@ -31,11 +37,13 @@ function Logger(level, filename) {
...
     level: logLevel,
     levels: javaLogLevels.levels,
+    colors: javaLogLevels.colors,
     transports: [
       new Winston.transports.Console({
         timestamp: true,
         json: Config.get('log:json'),
-        stringify: Config.get('log:json')
+        stringify: Config.get('log:json'),
+        colorize: !Config.get('log.json')
```
erran-r7: @davepgreene shouldn't this be `log:json`?

dgreene-r7: Good catch. Oddly it still works. Wonder if nconf does both colon and dot notation.

dgreene-r7:
@fmitchell-r7 per our conversation yesterday, I'm going to add tests to cover the `/properties` endpoints. I have a few PRs related to updating linting rules and dependencies that I want to land before I do that work, so I'm going to merge this, submit those PRs, then write the tests.

PR #231, opened 2017-03-28
Reduce Propsd user privileges

### LINKED ISSUE #230: Don't allow Propsd user too many privileges
Currently the Propsd user can write to the Propsd installation and configuration directory. We shouldn't do this.

dgreene-r7:
This PR changes the desired ownership of the Propsd installation and conf dir so that it's owned by the installing user (e.g. `root`) but still run by the Propsd user.

This resolves #230.


fmitchell-r7:
Does the [Upstart script](https://github.com/rapid7/propsd/blob/master/cookbook/templates/default/upstart.conf.erb#L9-L10) still work after this change? It's using `setuid` and `setgid` to run as the "propsd" user.

dgreene-r7:
Don't see why it wouldn't. This isn't removing the Propsd user or group, it's simply making sure that the Propsd user can't write to the application or configuration directory.

PR #237, opened 2017-04-11
added region to payload in tokendTransformer and relevant test case

dcooper-r7:
This enables support for KMS region specification in tokend.


test/tokend-transformer.js
```diff
@@ -139,7 +139,8 @@ describe('TokendTransformer', function() {
         $tokend: {
           type: 'kms',
           resource: '/v1/kms/decrypt',
-          ciphertext: 'gbbe'
+          ciphertext: 'gbbe',
+          region: 'eu-central-1'
```
fmitchell-r7: Can you split this into two tests, one that includes the region with the request and one that doesn't? The idea is that if we accidentally delete the newly added region handling code from `lib/transformers/tokend.js`, one of the tests should break.

dgreene-r7: I'd argue that another test case doesn't necessarily illustrate anything if we're not requiring the region with the request. Remember that [Tokend sets a default value](https://github.com/rapid7/tokend/blob/master/lib/providers/kms.js#L21-L28) if there's no `region` in the request.

fmitchell-r7: Tests ensure the code in `lib/transformers/tokend.js` does the right thing. For example, if you replace that code with this
```
if (info.get('region') && info.get('region') !== '') {
  payload.region = info.get('regⅰon');
}
```
all the test (as written) will pass. But you'll always end up using the default region in Tokend, no matter what region is defined in the Propsd property.

Depending on your font settings, it may or may not be obvious that "i" was replaced with [U+2170](http://graphemica.com/%E2%85%B0), but failing tests will catch that something's wrong.

PR #242, opened 2017-05-16
Feature: metadata source gets ASG info

### LINKED ISSUE #126: Metadata Source should get Auto Scaling Group info
Collecting ASG membership info doesn't fall under the basic Metadata source operation but being able to create an ASG and bring up canary nodes that share the same configuration is a really nice use case.

This work will involve additional async operations which may cause metadata resolution to take slightly longer so we should keep an eye on performance in the wild before we merge any work on this ticket to master.

dgreene-r7:
This PR adds functionality to the `Metadata` source to retrieve information about the instance's auto-scaling group. This resolves #126.

### DIFF
--- a/lib/source/metadata.js
+++ b/lib/source/metadata.js
@@ -99,6 +99,8 @@ class Metadata extends Source.Polling(Parser) { // eslint-disable-line new-cap
           p = new Promise((resolve, reject) => {
             (new Aws.AutoScaling({region})).describeAutoScalingInstances({InstanceIds: [instanceId]}, (err, d) => {
               if (err) {
+                Log.log('ERROR', err);
+
                 return reject(err);
               }
               resolve(d);
--- a/lib/source/metadata.js
+++ b/lib/source/metadata.js
@@ -87,24 +87,59 @@ class Metadata extends Source.Polling(Parser) { // eslint-disable-line new-cap
           return callback(err);
         }
 
-        // Detect change by hashing the fetched data
-        const hash = Crypto.createHash('sha1');
-        const paths = Object.keys(data);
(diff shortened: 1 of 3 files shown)


lib/source/metadata.js
```diff
@@ -87,24 +87,59 @@ class Metadata extends Source.Polling(Parser) { // eslint-disable-line new-cap
...
+              Log.log('WARN', `Instance id ${instanceId} is in multiple auto-scaling groups`, asg);
+            }
+
+            // Check to see if an instance is actually part of an ASG
+            if (asg.length !== 0) {
+              data['auto-scaling-group'] = asg[0];
+            }
+
+            return data;
+          }).catch(() => data);
```
fmitchell-r7: Is this `catch` going to swallow the error that a failed `describeAutoScalingInstances` call might throw?

dgreene-r7: Yes. If a promise fails it looks through all of its chained functions and finds the next `catch`. If it succeeds after that (i.e. when we return `data`) the next `then` will continue the execution.

fmitchell-r7: Today I learned something new about promises. Thanks!

In that case, do we want to log the error here and then continue?

erran-r7:
Would the `auto-scaling-group` be available for indexing/string templating in the index.yml? If so would we be able to handle ec2 tags in the same way for #243? 🎉

dgreene-r7:
> Would the auto-scaling-group be available for indexing/string templating in the index.yml? If so would we be able to handle ec2 tags in the same way for #243? 🎉

That's the goal!

PR #244, opened 2017-05-17
Support {{ instance:iam-role }} for indexing

erran-r7:
Context: One of our services is going to start consuming from different queues/buckets when given "role" of "on demand" instead of the default role. This lines up fairly closely with the instance profile/role since we want to only grant access based on those.

We can use the instance profile/role for separating services.

Instance Tags would be nice but this is very straight forward first step.

### DIFF
--- a/lib/source/metadata/parser.js
+++ b/lib/source/metadata/parser.js
@@ -116,6 +116,7 @@ Parser.mappings = {
       return;
     }
 
+    properties['iam-role'] = Path.basename(roles[0]);
     const credentials = JSON.parse(metadata[roles[0]]);
 
     if (!properties.credentials) {
--- a/test/data/s3/index.json
+++ b/test/data/s3/index.json
@@ -23,6 +23,13 @@
       }
     },
     {
+      "name": "iam-role",
+      "type": "s3",
+      "parameters": {
+        "path": "role/{{ instance:iam-role }}.json"
+      }
+    },
+    {
       "name": "consul",
       "type": "consul"
     }
--- a/test/data/s3/role/fake-fake.json
+++ b/test/data/s3/role/fake-fake.json
@@ -1,0 +1,6 @@
+{
+  "version": 1.0,
+  "properties": {
+    "only.set.on.fake-fake": true
+  }
+}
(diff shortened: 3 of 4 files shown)


test/data/metadata-paths.json
```diff
@@ -40,6 +40,7 @@
   "/latest/meta-data/services/partition": "aws",
   "/latest/meta-data/services/domain": "amazonaws.com",
   "/latest/meta-data/network/interfaces/macs/": "0e:9c:a1:fe:2f:ef/",
+  "/latest/meta-data/iam/security-credentials": "fake-fake",
```
erran-r7: This is already defined above.

test/data/metadata-values.json
```diff
@@ -26,6 +26,7 @@
   "meta-data/local-hostname": "ip-10-196-24-63.ec2.internal",
   "meta-data/instance-type": "t2.small",
   "meta-data/instance-id": "i-aaaf2d1a",
+  "meta-data/iam/security-credentials": "fake-fake",
```
erran-r7: The spec is failing saying this was in expected but not actual. It may not be necessary here. Adding the expectation to the spec should be enough.

erran-r7:
I've also validated this change by adding the `iam-role` assignment into the metadata parser and restarting propsd. I'm not sure how this behaves on an instance. For instances without a role we perform an early return as we previously did for credentials.

lib/source/metadata/parser.js
```diff
@@ -116,6 +116,7 @@ Parser.mappings = {
       return;
     }
 
+    properties['iam-role'] = roles[0].substr(path.length);
```
dgreene-r7: I'm hesitant to use `String.substr()` here where `String.replace()` is more appropriate for the operation and more obvious in its action.

dgreene-r7:
Also please rebase :)

lib/source/metadata/parser.js
erran-r7: @dgreene-r7 that looks so much better. 😍 The `substr` thing felt like a hack and I knew there was a more clear way to do it. Just my lack of touching javascript in a long time I guess.

erran-r7:
> Also please rebase :)

Rebased in d27ba65.

lib/source/metadata/parser.js
erran-r7: Opt'd for `Path.basename` in d27ba65. We use Path's join in other places when interacting with meta-data keys.

PR #245, opened 2017-05-19
Fix: Consul source should not be loaded by default

### LINKED ISSUE #223: Consul source should not be loaded by default
The Consul source loads automatically and doesn't respect the presence of a `consul` source in the index. 

In order to support an environment that doesn't use Consul without the current behavior of a constantly erroring plugin, we should make this source load only if it's included in the index.

dgreene-r7:
This PR removes the Consul source from being loaded by default as a `dynamic` layer. Instead, it is loaded as a regular source. In order to maintain backwards compatibility, the set of Consul properties is namespaced in the same way that it was before this change.

Resolves #223.

### DIFF
--- a/bin/server.js
+++ b/bin/server.js
@@ -21,7 +21,6 @@ const Logger = require('../lib/logger');
 
 const Properties = require('../lib/properties');
 const Sources = require('../lib/sources');
-const Consul = require('../lib/source/consul');
 const Metadata = require('../lib/source/metadata');
 const Tags = require('../lib/source/tags');
 const S3 = require('../lib/source/s3');
@@ -57,7 +56,6 @@ const properties = new Properties();
 const sources = new Sources(properties);
 
 // Add metadata and some statics
-properties.dynamic(new Consul('consul', Config.get('consul')), 'consul');
 properties.dynamic(new Metadata(Config.get('metadata')), 'instance');
 properties.dynamic(new Tags(Config.get('metadata')), 'instance:tags');
 properties.static(Config.get('properties'));
--- a/lib/source/consul.js
+++ b/lib/source/consul.js
@@ -24,9 +24,9 @@ class Consul extends Source.Polling(Parser) { // eslint-disable-line new-cap
    */
   constructor(name, opts) {
     const options = Object.assign({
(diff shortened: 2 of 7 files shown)


bin/server.js
```diff
@@ -56,7 +55,6 @@ const properties = new Properties();
 const sources = new Sources(properties);
 
 // Add metadata and some statics
-properties.dynamic(new Consul('consul', Config.get('consul')), 'consul');
```
fmitchell-r7: When do we create the Consul source? I can't find any other references to `new Consul()` where the config would get loaded.

dgreene-r7: It has to be specified in the index like [this](https://github.com/rapid7/propsd/blob/781799eec8343f124dea0f7323fbb365ebac8fd5/test/data/s3/index.json#L32-L35).

lib/source/consul.js
```diff
@@ -24,9 +24,9 @@ class Consul extends Source.Polling(Parser) { // eslint-disable-line new-cap
    */
   constructor(name, opts) {
     const options = Object.assign({
-      host: Consul.DEFAULT_ADDRESS,
-      port: Consul.DEFAULT_PORT,
-      secure: false
+      host: Config.get('consul:host'),
+      port: Config.get('consul:port'),
+      secure: Config.get('consul:secure')
     }, opts);
```
fmitchell-r7: Do you think it's worth doing this?
```javascript
const options = Object.assign({
  host: Consul.DEFAULT_ADDRESS,
  port: Consul.DEFAULT_PORT,
  secure: false
}, Config.get('consul'), opts);
```
That way we get defaults in case users leave something out of the config. Plus, we still get to overwrite options in the constructor for testing.

dgreene-r7: If you look in the [default config](https://github.com/rapid7/propsd/blob/781799eec8343f124dea0f7323fbb365ebac8fd5/config/defaults.json#L15-L19) we already specify some Consul options. Imo better to not duplicate these things.

dgreene-r7:
Because this PR introduces breaking changes that would necessitate a major release, I want to land #246 first.

PR #246, opened 2017-05-25
Feature: EC2 tags source

### LINKED ISSUE #243: Add EC2Tags source
@fmitchell-r7 [made a great suggestion](https://github.com/rapid7/propsd/issues/126#issuecomment-296213400) in #126:
> Alternatively, we could call [describeTags](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#describeTags-property), which would give us both the auto scaling group name (via the `aws:autoscaling:groupName` tag) as well as any custom user defined tags. Tags could be referenced in the index file as `instance:tag:{key}:{value}`. We would need a way to deal with spaces and other odd characters in tags.

We should implement this as a `Polling` source and load it as a dynamic property after EC2 metadata gets loaded.

dgreene-r7:
This PR adds support for using instance tags returned from the EC2 API as dynamic variables in indices. If Propsd isn't running on an EC2 instance or doesn't have access to the Tags API the source will fail silently in the same way as the EC2 metadata service.

Per #243 tags can be referenced in the index file like so:
```json
{
  "name": "name-tag",
  "type": "s3",
  "parameters": {
    "path": "{{ instance:tags:SomeTag }}.json"
  }
}
```
The result is, assuming the tag `SomeTag`'s value is `foobar`, that a file named `foobar.json` will be loaded as an S3 source.

The Tags source is implemented as a polling source.

This resolves #243.


lib/source/tags.js
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   * Fetch implementation for EC2 tags API
+   * @param {Function} callback
+   * @private
+   */
+  _fetch(callback) {
+    const path = `/${Metadata.version}/dynamic/instance-identity/document`;
+
+    new Promise((resolve) => {
```
fmitchell-r7: Do you need a `return` keyword here to return this promise?

dgreene-r7: Nope. The function that [calls _fetch()](https://github.com/rapid7/propsd/blob/master/lib/source/common.js#L333) doesn't do anything with a return value.

PR #249, opened 2017-06-07
Fix: empty properties are persisted as `null`.

dgreene-r7:
This PR fixes a regression introduced in 4e925e4 where empty properties were assigned null rather than removed from the properties object entirely.

### DIFF
--- a/lib/properties.js
+++ b/lib/properties.js
@@ -155,7 +155,7 @@ class Properties extends EventEmitter {
         if (!layer.namespace) {
           return merge(properties, layer.properties);
         }
-        properties[layer.namespace] = layer.properties;
+        properties[layer.namespace] = merge(properties[layer.namespace], layer.properties);
 
         return properties;
       }, {});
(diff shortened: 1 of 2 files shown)


lib/properties.js
```diff
@@ -111,29 +151,27 @@ class Properties extends EventEmitter {
...
+      const persistent = this.layers.reduce((properties, layer) => {
         if (!layer.namespace) {
-          return p.mergeDeep(Immutable.fromJS(layer.properties));
+          return merge(properties, layer.properties);
         }
-        const layerProps = Immutable.Map(layer.properties);
-
-        return p.setIn(layer.namespace.split(':'), layerProps);
-      }, new Immutable.Map());
+        properties[layer.namespace] = layer.properties;
```
fmitchell-r7: Does this need to be
```javascript
properties[layer.namespace] = merge(properties[layer.namespace], layer.properties);
```
to match the [original behavior](https://github.com/rapid7/propsd/commit/4e925e44ab0cf09e93a2f77089b2e8a3abc468a2?diff=split#diff-327455fa29663c58312adf5955056773L129)?

PR #253, opened 2017-06-16
Fix tag source is not nested

dgreene-r7:
This PR resolves an issue where the `Tag` source would not nest its data in the `instance` object but instead create a new object with key `instance:tags`.

### DIFF
--- a/lib/properties.js
+++ b/lib/properties.js
@@ -4,6 +4,7 @@ const Layer = require('./properties/layer');
 const View = require('./properties/view');
 const TokendTransformer = require('./transformers/tokend');
 const Immutable = require('immutable');
+const isPlainObject = require('lodash.isplainobject');
 
 /* eslint-disable eqeqeq */
 /**
@@ -16,11 +17,10 @@ const Immutable = require('immutable');
  */
 const merge = (destination, source) => {
   // Ensure that the destination value is an Object. `== null ` covers both null and undefined
-  const dest = (destination == null ||
-  Object.getPrototypeOf(destination) !== Object.prototype) ? {} : destination;
+  const dest = isPlainObject(destination) ? destination : {};
 
   // Only merge source if it's an Object. `== null ` covers both null and undefined
-  if (source == null || Object.getPrototypeOf(source) !== Object.prototype) {
+  if (!isPlainObject(source)) {
     return dest;
   }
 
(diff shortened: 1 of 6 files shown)


lib/properties.js
```diff
@@ -16,11 +17,10 @@ const Immutable = require('immutable');
  */
 const merge = (destination, source) => {
   // Ensure that the destination value is an Object. `== null ` covers both null and undefined
-  const dest = (destination == null ||
-  Object.getPrototypeOf(destination) !== Object.prototype) ? {} : destination;
+  const dest = isPlainObject(destination) ? destination : {};
 
   // Only merge source if it's an Object. `== null ` covers both null and undefined
```
fmitchell-r7: I don't think this comment applies any more, since we're using `isPlainObject` instead of `source == null`.

lib/properties.js
```diff
@@ -46,6 +46,28 @@ const merge = (destination, source) => {
...
+ * @param {Array<String>} namespaceArray An array of keys (namespaces) to traverse
+ * @param {Object} source The source that properties are merged from
+ * @return {Object}
+ */
+const recusiveNamespaceMerge = (destination, namespaceArray, source) => {
+  const nextNamespace = namespaceArray.shift();
+  const dest = isPlainObject(destination) ? destination : {};
+
+  if (namespaceArray.length) {
+    dest[nextNamespace] = recusiveNamespaceMerge(dest[nextNamespace] || {}, namespaceArray, source);
```
fmitchell-r7: Can we eliminate the `|| {}` here? The `isPlainObject()` call above should handle it.

lib/transformers/tokend.js
```diff
@@ -14,7 +15,7 @@ function collectTransformables(properties, keyPath) {
   let results = [];
 
   // Ensure we're walking an Object. `== null` covers both null and undefined.
```
fmitchell-r7: I don't think this comment applies any more, since we're using `isPlainObject` instead of `source == null`.

lib/transformers/tokend.js
```diff
@@ -29,7 +30,7 @@ function collectTransformables(properties, keyPath) {
     const value = properties[key];
 
     // Don't walk anything that's not an Object. `!= null` covers both null and undefined.
```
fmitchell-r7: I don't think this comment applies any more, since we're using `isPlainObject` instead of `source == null`.

lib/properties.js
dgreene-r7: It doesn't. Line 57 can either give you an object with indeterminate keys or an empty object. The check on line 60 makes sure that indexing into that object gives you something rather than undefined.

PR #254, opened 2017-07-11
Implement babel

dgreene-r7:
This PR adds Babel to the build chain for Propsd.

When this PR is merged we're free to begin migrating the codebase to modern Javascript with the option of shifting our transpilation target as we go. This allows us to decouple our development process from the way we package the app.


package.json
```diff
@@ -51,6 +50,12 @@
   },
   "dependencies": {
     "aws-sdk": "~2.2.28",
+    "babel-cli": "^6.24.1",
```
fmitchell-r7: Can the `babel-*` packages be development dependencies, or are they needed at run time? If they're only needed to run the transpile process (and aren't needed at run time) can we do something like
```bash
npm install
npm run transpile
npm prune --production
``` 
before copying the ./node_modules folder for packaging?

dgreene-r7: They're needed at runtime because we're using the transform runtime plugin. It's not really that big of a deal, though. Package size isn't increased by much.

dgreene-r7:
As an FYI, the branch build of Nock is being tracked in https://github.com/node-nock/nock/pull/929. We should make sure that when it's merged we revert back to an official release.

package.json
```diff
@@ -2,19 +2,17 @@
...
+    "clean": "rimraf dist",
+    "version": "./test/bin/version.js",
+    "transpile": "npm run version && babel src --out-dir dist --copy-files",
+    "test": "npm run transpile && mocha",
+    "lint": "eslint src/**/*.js test/**/*.js",
+    "coverage": "npm run transpile && istanbul cover _mocha -- -R spec",
     "report-coverage": "npm run coverage && cat ./coverage/lcov.info | coveralls && rm -rf ./coverage",
-    "metadata-server": "./test/bin/metadata-server",
-    "s3-server": "./test/bin/s3-server.js",
+    "metadata-server": "nodemon ./test/bin/metadata-server",
```
fmitchell-r7: Can we change the command to `nodemon ./test/bin/metadata-server --` so arguments passed on the command line to `npm run metadata-server` are passed to the metadata-server instead of nodemon?

package.json
```diff
@@ -2,19 +2,17 @@
...
+    "s3-server": "nodemon ./test/bin/s3-server.js",
```
fmitchell-r7: Can we change the command to `nodemon ./test/bin/s3-server.js --` so arguments passed on the command line to `npm run s3-server` are passed to the s3-server instead of nodemon?

package.json
```diff
@@ -2,19 +2,17 @@
...
-  "directories": {
-    "doc": "docs",
-    "bin": "bin"
-  },
   "scripts": {
-    "start": "./bin/server.js",
-    "test": "mocha",
-    "lint": "eslint bin/** lib/** test/**",
-    "coverage": "istanbul cover _mocha -- -R spec",
+    "start": "nodemon ./src/bin/server.js -r babel-require",
```
fmitchell-r7: Can we change the command to `nodemon ./src/bin/server.js -r babel-require --` so arguments passed on the command line to `npm run start` are passed to the Propsd server instead of nodemon?

package.json
```diff
@@ -2,19 +2,17 @@
...
+    "transpile": "npm run version && babel src --out-dir dist --copy-files",
+    "test": "npm run transpile && mocha",
+    "lint": "eslint src/**/*.js test/**/*.js",
+    "coverage": "npm run transpile && istanbul cover _mocha -- -R spec",
     "report-coverage": "npm run coverage && cat ./coverage/lcov.info | coveralls && rm -rf ./coverage",
-    "metadata-server": "./test/bin/metadata-server",
-    "s3-server": "./test/bin/s3-server.js",
+    "metadata-server": "nodemon ./test/bin/metadata-server",
+    "s3-server": "nodemon ./test/bin/s3-server.js",
     "dev-server": "npm run metadata-server & npm run s3-server -- -d test/data/s3 & npm start -- -c config/dev.json"
```
fmitchell-r7: Do we want to change the "config/dev.json" path to "src/config/dev.json"?

package.json
dgreene-r7: I did a quick audit and I'm going to prune down the dependencies we're shipping. Per https://babeljs.io/docs/plugins/transform-runtime/:

>NOTE - Production vs. development dependencies

> In most cases, you should install babel-plugin-transform-runtime as a development dependency (with --save-dev) and babel-runtime as a production dependency (with --save).

> The transformation plugin is typically used only in development, but the runtime itself will be depended on by your deployed/published code. See the examples below for more details.

Moving the `babel-*` packages to `devDependencies` and adding `babel-runtime` to regular `dependencies` cut the extracted size from 70mb to 30mb and the package size from 8.8mb to 4.2mb.

dgreene-r7:
The fix for the Nock bug has been merged to master but a new release hasn't been cut. The next step in them fixing their tests is https://github.com/node-nock/nock/issues/947.

PR #257, opened 2017-10-11
Fix: Add Tokend caching

dgreene-r7:
This PR caches responses from Tokend so we don't bust KMS API limits by doing decrypt operations  on every secret every time we hit the `/conqueso` or `/properties` endpoint.

### DIFF
--- a/src/config/defaults.json
+++ b/src/config/defaults.json
@@ -20,7 +20,8 @@
   "tokend": {
     "host": "127.0.0.1",
     "port": 4500,
-    "interval": 300000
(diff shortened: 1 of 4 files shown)


fmitchell-r7:
Is it worth putting a limit on the size of the cache and/or clearing it out at some point? I'm thinking mostly, "What happens when secrets change and we have old secrets in the cache that will never by accessed again?"

dgreene-r7:
Yeah, that's a really good suggestion. I was thinking about doing this but completely forgot.

PR #258, opened 2018-03-30
Refactor metadata path traversal

davepgreene:
One of the major difficulties we've had with Propsd is handling legacy complexity. This PR refactors `metadata/util`'s `traverse` function. Previously it relied heavily on `each` which was a bizarre state machine implementation that ultimately isn't required as the "parallelism" it touted wasn't actually used.

Now, instead of using `each`, we operate on an array in a `while` loop, `shift` the first element off, process it, and continue until the array is empty. In the case of additional paths we push them onto the end of the array. Because we're using a `while` loop and `shift`ing we don't have to worry about modifying the array during iteration (because we're not actually iterating.)

The PR also bumps the version of the aws-sdk and aws-sdk-mock because they're seriously old and outdated. It also reorganizes the `Metadata._fetch` function for readability.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   },
   "homepage": "https://github.com/rapid7/propsd#readme",
   "devDependencies": {
-    "aws-sdk-mock": "~1.2.0",
+    "aws-sdk-mock": "^1.7.0",
     "babel-cli": "^6.24.1",
     "babel-plugin-add-module-exports": "^0.2.1",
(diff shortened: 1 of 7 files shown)


davepgreene:
cc: @tquetano-r7

dgreene-r7:
cc: @rapid7/platform-delivery-cambridge as well

dgreene-r7:
Good call. I'll update the PR when I get home tonight.

PR #262, opened 2019-05-13
PD-20473-store-previous-properties

aung-r7:
## Description:
This pull request is created to update propsd to continuously hold previously encrypted properties when a S3 / KMS failure occurs. We have experienced encrypted properties being set to `null` when we're unable to communicate to KMS.

## Note:
Context:
- The tokend transformer creates a cache that gets cleared every 5 mins (this is default). When an encrypted property is successfully decrypted from tokend, propsd stores the plaintext secret in this cache, after 5 mins this cache gets cleared and then propsd makes calls to tokend to decrypt again, and this process repeats every 5 mins.

Problem:
- If the cache gets cleared, and propsd makes a call to tokend to decrypt an encrypted property but tokend responds back with a non-200 status code (saying it's unable to decrypt the property), what value should I set for this encrypted property? In the end, it gets set to null.

Implementation:
- The idea here is to not clear the cache on a time interval, and instead update it accordingly.

## Testing:
I tested this by spinning up a dummy instance, and modified the code to have this implementation. Below are the test cases I have experimented with and the steps that I used to verify the output.

Test cases:
- S3 Failures
    1. Logged onto the dummy instance, made a request to the `v1/conqueso` endpoint and was able to receive properties successfully (including encrypted properties)
    2. Removed the IAM inline policy (allowing access to the `config.core-0` bucket) that is attached to the IAM role of this dummy instance
    3. While tailing the propsd logs, we start to see logs with errors saying `Access Denied` about accessing the `config.core-0` bucket.
    4. While on the instance, made a request to the `v1/conqueso` endpoint and was able to receive properties successfully (including encrypted properties)
    5. Updated a property
    6. Added back the IAM inline policy in step 2
    7. While on the instance, made a request to the `v1/conqueso` endpoint and was able to receive properties successfully (including encrypted properties), and validated that the property I changed in step 5 was updated with the new value

- KMS Failures

    1. Logged onto the dummy instance, made a request to the `v1/conqueso` endpoint and was able to receive properties successfully (including encrypted properties)
    2. Modified tokend response to return a status code of 500 (failed to decrypt successfully), restart the tokend process
    3. Updated a encrypted property, validated that propsd made a request tokend to decrypt newly updated encrypted property and failed with a status code of 500
    4. While on the instance, made a request to the `v1/conqueso` endpoint and was able to receive properties successfully (including encrypted properties), and validated that the encrypted property I modified in step 3 is still providing it's old value

### DIFF
--- a/src/lib/transformers/tokend-client.js
+++ b/src/lib/transformers/tokend-client.js
@@ -259,6 +259,11 @@ class TokendClient extends Source.Polling(TokendParser) {
         } catch (e) {
           return callback(e, null);
         }
+
+        if (res.statusCode !== 200) {
+          callback(new Error("ERROR: Received a non 200 status code from tokend"), secret.error.message);
+        }
+
         callback(null, secret);
       });
     });
--- a/src/lib/transformers/tokend-client.js
+++ b/src/lib/transformers/tokend-client.js
@@ -261,7 +261,7 @@ class TokendClient extends Source.Polling(TokendParser) {
         }
 
         if (res.statusCode !== 200) {
-          callback(new Error("ERROR: Received a non 200 status code from tokend"), secret.error.message);
+          callback(new Error(`ERROR: Received '${secret.error.message}' from tokend while trying to decrypt a secret`), null);
         }
 
         callback(null, secret);
--- a/src/lib/transformers/tokend.js
+++ b/src/lib/transformers/tokend.js
@@ -73,12 +73,7 @@ class TokendTransformer {
     const seenProperties = [];
     const promises = collectTransformables(properties, []).map((info) => {
       const keyPath = info.get('keyPath');
-      if (!Array.isArray(keyPath) || keyPath.length !== 1) {
-        Log.log('WARN', 'keyPath is not an array or  does not have a length of 1, this could cause errors in the property set')
-        return Promise.reject('keyPath is not an array or  does not have a length of 1')
-      }
-
-      const propertyName = keyPath[0];
+      const propertyName = keyPath.join(".");
       seenProperties.push(propertyName);
 
       const signature = crypto
--- a/src/lib/transformers/tokend.js
+++ b/src/lib/transformers/tokend.js
@@ -74,6 +74,7 @@ class TokendTransformer {
     const promises = collectTransformables(properties, []).map((info) => {
       const keyPath = info.get('keyPath');
       const propertyName = keyPath[0];
+      seenProperties.push(propertyName);
 
       const signature = crypto
         .createHash('sha1')
@@ -154,13 +155,11 @@ class TokendTransformer {
           plaintext: data.plaintext
         };
 
-        seenProperties.push(propertyName);
-
         return Promise.resolve(Immutable.Map().setIn(keyPath, data.plaintext));
       }).catch((err) => {
         Log.log('WARN', err);
         this._client.clearCacheAtKey(method, requestId);
-        if (this._cache.hasOwnProperty(propertyName) && this._cache[propertyName].signature === signature) {
+        if (this._cache.hasOwnProperty(propertyName)) {
           return Promise.resolve(Immutable.Map().setIn(keyPath, this._cache[propertyName].plaintext));
         }
 
--- a/src/lib/transformers/tokend.js
+++ b/src/lib/transformers/tokend.js
@@ -73,8 +73,9 @@ class TokendTransformer {
     const seenProperties = [];
     const promises = collectTransformables(properties, []).map((info) => {
       const keyPath = info.get('keyPath');
-      if (keyPath.length !== 1) {
-        Log.log('WARN', 'keyPath does not have a length of 1, this could cause errors in the property set')
+      if (!Array.isArray(keyPath) || keyPath.length !== 1) {
+        Log.log('WARN', 'keyPath is not an array or  does not have a length of 1, this could cause errors in the property set')
+        return Promise.reject('keyPath is not an array or  does not have a length of 1')
       }
 
       const propertyName = keyPath[0];
@@ -126,7 +127,6 @@ class TokendTransformer {
             payload.datakey = info.get('datakey');
           }
           method = 'POST';
-
           resolver = this._client.post(info.get('resource'), payload);
           break;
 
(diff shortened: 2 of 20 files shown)


aung-r7:
I want to get some initial thoughts on this before I start writing tests. Let me know what you think!

src/lib/transformers/tokend-client.js
```diff
@@ -259,6 +259,11 @@ class TokendClient extends Source.Polling(TokendParser) {
         } catch (e) {
           return callback(e, null);
         }
+
+        if (res.statusCode !== 200) {
+          callback(new Error("ERROR: Received a non 200 status code from tokend"), secret.error.message);
```
aung-r7: Poopie, this should probably be something along the lines of "ERROR: Tokend failed to decrypt a secret"

dgreene-r7: I'm not sure you want to pass the error message as the `data` argument to the callback as it'll be thrown away when the error is handled. See https://github.com/rapid7/propsd/pull/262/files#diff-b3189f329f8903f3731284636ec9e9f6R150.

A better option might be to include the error message from Tokend in the `Error` that's being passed as the first argument.

src/lib/transformers/tokend.js
```diff
@@ -171,12 +149,18 @@ class TokendTransformer {
...
 
         return Promise.resolve(Immutable.Map().setIn(keyPath, data.plaintext));
       }).catch((err) => {
         Log.log('WARN', err);
         this._client.clearCacheAtKey(method, requestId);
+        if (this._cache.hasOwnProperty(propertyName)) {
+          return Promise.resolve(Immutable.Map().setIn(keyPath, this._cache[propertyName].plaintext));
+        }
 
         return Promise.resolve(Immutable.Map().setIn(keyPath, null));
```
dgreene-r7: Might be worth adding some logging to surface when there's an error from Tokend and the property doesn't exist in the cache.

src/lib/transformers/tokend.js
```diff
@@ -96,22 +70,26 @@ class TokendTransformer {
    * @return {Promise<Properties>} properties that had $tokend values transformed
    */
   transform(properties) {
+    const seenProperties = [];
     const promises = collectTransformables(properties, []).map((info) => {
       const keyPath = info.get('keyPath');
+      const propertyName = keyPath[0];
```
dgreene-r7: Might be worth it to throw a length check and some logging around `keyPath`. The way we're using Propsd right now it _should_ only have 1 element but it'd be worth surfacing whether this is always the case.

src/lib/transformers/tokend.js
```diff
@@ -189,6 +173,12 @@ class TokendTransformer {
         transformedProperties = transformedProperties.mergeDeep(value);
       });
 
+      Object.keys(this._cache)
```
dgreene-r7: This is going to sound dumb but can you put some comments around this ejection strategy? 

The approach should be fine for the size of the property-set we have but I can totally see this chugging if the cache gets large enough and it would be useful to capture the reasoning for the approach now in case we need to revisit down the road.

src/lib/transformers/tokend.js
```diff
@@ -189,6 +173,12 @@ class TokendTransformer {
...
+        .filter((propertyName) => seenProperties.indexOf(propertyName) === -1)
+        .forEach((entry) => {
```
dgreene-r7: You can probably consolidate this line and the previous one within a single `forEach` (or even a regular `for` loop) with an `if` statement. Both `filter` and `forEach` are `O(n)` so you're iterating over the full list, then a subset just to remove values from the array when the time spent in each iteration grows linearly based on the number of properties in the property-set.

dgreene-r7:
@aung-r7 you'll need to update tests regardless of the approach you take. When you do, can you take a look at that failing ASG-related test? If you have questions feel free to ask :)

src/lib/transformers/tokend-client.js
aung-r7: All set

src/lib/transformers/tokend.js
aung-r7: All set, added a Log.log(`Warn`) message after the if block

src/lib/transformers/tokend.js
aung-r7: All set, add a check if the keyPath length is not equal to 1 and log if that is the case

src/lib/transformers/tokend.js
aung-r7: It is not dumb at all! I added a comment, let me know if the comment is good enough.

src/lib/transformers/tokend.js
aung-r7: All set

src/lib/transformers/tokend.js
```diff
@@ -96,22 +70,30 @@ class TokendTransformer {
...
    */
   transform(properties) {
+    const seenProperties = [];
     const promises = collectTransformables(properties, []).map((info) => {
       const keyPath = info.get('keyPath');
+      if (keyPath.length !== 1) {
+        Log.log('WARN', 'keyPath does not have a length of 1, this could cause errors in the property set')
+      }
+
+      const propertyName = keyPath[0];
```
dgreene-r7: What if `keyPath` is `null`, `undefined`, or something that's not an array with 1 element?

aung-r7: ahh good catch, will update the if condition above to check, maybe we should throw an error if for this? When I talked to you before I believe we said logging should be fine?

dgreene-r7: I think the `!== 1` might be a little aggressive. There may be a valid case where it returns an array with a length of 2 but we want the first element regardless.

My thought is definitely `reject` if `keyPath` isn't an array or is an array with `length == 0`. Otherwise, if `length > 1`, warn.

I feel like that covers all possible cases. What do you think?

aung-r7: yeah, my `!== 1` is too aggressive and it failed some tests. I am going to do a little bit of debugging / comparison between master and this branch based on these tests just to make sure I am not missing anything drastic.

dgreene-r7:
Nice job on this!

PR #278, opened 2020-03-24
Updating the propsd health-check to reflect initialising state

snorlaX-sleeps:
This change is to modify the propsd health-check.
Currently, the health-check will return with a `200` unless a source or index is in `WARNING` or `ERROR` state
However, we have seen scenarios where the health-check will return `200` when sources are in an `INITIALIZING` or `CREATING` state
These states should return a not ready Status Code
Having some sources in an error state should not stop the service from being used, however having all in error should.

Worked on with @davepgreene

### DIFF
--- a/src/lib/sources.js
+++ b/src/lib/sources.js
@@ -167,6 +167,11 @@ class Sources extends EventEmitter {
       // TODO This logic is fairly ham-fisted right now. It'll work because nothing
       // is setting `state` to WARNING at the moment. When we do start supporting a
       // WARNING state, this will have to become aware that OK < WARING < ERROR.
+      if (!source.ready) {
+        object.code = STATUS_CODES.SERVICE_UNAVAILABLE;
+        object.status = source.state;
+      }
+
       if (!source.ok) {
         object.code = STATUS_CODES.INTERNAL_SERVER_ERROR;
         object.status = source.state;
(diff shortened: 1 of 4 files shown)


src/lib/sources.js
```diff
@@ -171,14 +176,24 @@ class Sources extends EventEmitter {
     });
 
     object.sources = this.properties.sources.map((source) => {
-      if (!source.ok) {
```
dgreene-r7: I think we want to apply this same logic to indices as well.

test/core-api-v1.js
```diff
@@ -4,20 +4,23 @@ const request = require('supertest');
 const Properties = require('../dist/lib/properties');
 const Sources = require('../dist/lib/sources');
 const S3 = require('../dist/lib/source/s3');
+global.Config = require('nconf')
```
dgreene-r7:
```suggestion
require('./lib/helpers');
```

test/sources.js
```diff
@@ -355,38 +355,120 @@ describe('Sources', function() {
   });
 
   describe('Health', function() {
-    const stubs = setUp();
+    this.timeout(5000);
```
dgreene-r7: We should avoid setting test timeouts as tweaking this can mask actual timing issues in our code.

aRobinson-R7: will look

aRobinson-R7: As mentioned, the previous section of the tests (`initialization`) already has a timeout value, which indicates this is something to do with the initialization process, which has always been in

PR #279, opened 2020-03-27
Feature update status code based on property set state

dgreene-r7:
This PR:
1. Bumps a bunch of our test dependencies with corresponding changes.
1. Updates tests to deprecate the use of `chai`, `chai-as-promised`, and Mocha's assertion library in favor of standardizing on Should.js.
1. Renames a few key methods to ensure that they don't conflict with builtins (e.g. `static`).
1. Cleans up a whole bunch of linting issues.
1. Migrates some utils into a new file.


src/lib/sources.js
```diff
@@ -153,53 +154,63 @@ class Sources extends EventEmitter {
...
+      obj.status = u.status().state;
 
-    if (object.sources.length === source_state.count) {
-      object.code = STATUS_CODES.INTERNAL_SERVER_ERROR;
-      object.status = source_state.source[source_state.count - 1];
+      return obj;
     }
 
-    return object;
+    return obj;
```
aRobinson-R7: This method seems a lot neater - more JS way of doing things?
Setting the states to begin, returning once one error code gets updated?

package.json
```diff
@@ -25,53 +25,55 @@
...
+    "@babel/plugin-proposal-numeric-separator": "7.8.3",
+    "@babel/plugin-proposal-throw-expressions": "7.8.3",
+    "@babel/plugin-transform-runtime": "7.9.0",
+    "@babel/preset-env": "7.9.0",
+    "@babel/preset-stage-2": "7.8.3",
+    "aws-sdk-mock": "5.1.0",
+    "babel-eslint": "10.1.0",
+    "chokidar": "3.3.1",
+    "coveralls": "3.0.11",
+    "eslint": "6.5.1",
```
aRobinson-R7: eslint-config-rapid7@2.11.1" has incorrect peer dependency "eslint@^3.0.0 || ^4.0.0

dgreene-r7: Yeah, it's a known issue but doesn't affect how linting works. This is the only combination of updated `eslint` and `eslint-config-rapid7` that I was able to get working without errors. This is due to an issue in [rapid7/eslint-plugin/rapid7:sort-object-keys](https://github.com/rapid7/eslint-plugin-rapid7/blob/master/lib/rules/sort-object-keys.js#L5-L9) where it attempts to import an eslint lib that's no longer part of the project.

src/lib/sources.js
dgreene-r7: I'm not jazzy about the implementation but it captures our logic well. I'm definitely open to feedback about the actual code.

src/lib/sources.js
```diff
@@ -153,53 +154,63 @@ class Sources extends EventEmitter {
...
-    });
+      return obj;
+    }
 
-    object.sources = this.properties.sources.map((source) => {
-      if (!source.ready) {
-        object.code = STATUS_CODES.SERVICE_UNAVAILABLE;
-        object.status = source.state;
-      }
+    if (this.properties.sources.every(unhealthy)) {
```
aRobinson-R7, line 195: `every` does not work as intended here - probably safer with the current function
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

aRobinson-R7: possibly need to add a `!`

aRobinson-R7: Ignore this - there was an issue with my local git branch for this PR after the force-push fix that was done and it does work

dgreene-r7: You are right that there's an issue here. I need to do a sources length check because `every` returns true for empty arrays but no props because of no sources is still a "healthy" state even if it's not an expected use.

